PR #4, opened 2014-10-10
Pass through status in plc_tag_status

ericorange:
Previously only returned PLCTAG_ERR_NOT_IMPLEMENTED
in plc_tag_status if the tag->vtable or tag->vtable->status
were false. This resulted in many different errors being
hidden behind one error code.

Added code where when the above condition occurs, the tag->status
will be passed through if it is non-zero.

### DIFF
--- a/lib/libplctag_tag.c
+++ b/lib/libplctag_tag.c
@@ -267,6 +267,11 @@ LIB_EXPORT int plc_tag_status(plc_tag tag)
         return PLCTAG_ERR_NULL_PTR;
 
     if(!tag->vtable || !tag->vtable->status) {
+        if(tag->status) {
+            pdebug(tag->debug, "tag status not ok!");
+            return tag->status;
+        }
+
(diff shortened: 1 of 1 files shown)


kyle-github:
Hi @ericorange,

It has become clear over the last half-year or so that the LGPL license is becoming a problem.  Development processes have changed with a move to standalone, statically linked executables.   The LGPL is much easier to work with when dynamically linking.   

I have done some research on alternate possibly licenses and it seems that the MPL 2.0 has a good balance of rights for users and coders that allows things like static linking much more easily.   I would like to relicense the library under both the LGPL and the MPL 2.0.

Would you consider relicensing your contributions under the same dual license?

Best,
Kyle

ericorange:
Yes. I do not have any issues with relicensing any of my contributions under the LGPL/MPL 2.0 dual license.

kyle-github:
Thanks!

PR #20, opened 2015-12-19
Various fixes/improvements/contributions

thatcadguy:
@kyle-github Figured I'd give you a PR before getting too much farther. Take a look at the commit messages, but basically just small fixes/improvements here and there.


examples/multithread_cached_read.c
```diff
@@ -143,7 +144,7 @@ void thread_func(void *data)
 				
 		end = time_ms();
 		
-		fprintf(stderr,"%ld Thread %d got result %d with return code %s in %dms\n",time_ms(),tid,value,decode_error(rc),(int)(end-start));
+		fprintf(stderr,"%" PRId64 " Thread %d got result %d with return code %s in %dms\n",time_ms(),tid,value,decode_error(rc),(int)(end-start));
```
thatcadguy: This is the format specifier for int64_t from inttypes.h, GCC was complaining about %ld

examples/Makefile
```diff
@@ -18,84 +18,21 @@
...
-
-multithread_plc5: multithread_plc5.c
-	$(CC) -o multithread_plc5 multithread_plc5.c $(CFLAGS) $(SHLIBS)
-
-multithread_plc5_dhp: multithread_plc5_dhp.c
-	$(CC) -o multithread_plc5_dhp multithread_plc5_dhp.c $(CFLAGS) $(SHLIBS)
-
-multithread_cached_read: multithread_cached_read.c
-	$(CC) -o multithread_cached_read multithread_cached_read.c $(CFLAGS) $(SHLIBS)
+%: %.c
```
thatcadguy: Everything was the same format so I used wildcard build rules here

contrib/Python/plctag/libplctag.py
```diff
@@ -79,7 +79,7 @@ def defineFloatFunc(name, args):
 # to load the compiled libplctag C library from the current environment
 # or LD_LIBRARY path.
 system = platform.system()
-library = "plctag.dll" if system == "Windows" else "libplctag.so"
+library = "libplctag.dll" if system == "Windows" else "libplctag.so"
```
thatcadguy: MSVC makes libplctag.dll

examples/tag_rw.c
```diff
@@ -89,6 +91,55 @@ void my_sleep(int seconds)
 #endif
 
 
+const char* decode_error(int rc)
```
thatcadguy: I think tag_rw is much easier to use with this added

examples/vs.mk
```diff
@@ -18,7 +18,7 @@
 #    USA
 
 
-CFLAGS=/Iutils /I..\lib /MTd /DWIN32=1 
+CFLAGS=/Iutils /I..\lib /MTd /DWIN32=1 /W3
```
thatcadguy: equivalent to GCC's -Wall, more specifically /W3 is supposed to be "production-level" warnings

lib/util/attr.h
```diff
@@ -29,9 +29,10 @@
 #define ATTR_H_
 
 
-
+typedef struct attr_entry_t *attr_entry;
 typedef struct attr_t *attr;
 
+attr_entry find_entry(attr a, const char *name);
```
thatcadguy: Clang complained about missing function prototypes here

PR #22, opened 2016-01-01
Initial CMake attempt + small tweaks

thatcadguy:
Put an initial try at a CMake script in contrib; got a maybe-uninitialized warning from GCC that I missed before; add optimization flags to makefiles.


kyle-github:
I am interested in the gcc warning fix, but the others are less firm.  

The optimization flags are of unclear value.  Are you seeing cases where the library is using a lot of CPU?  I think the argument for -Os (small size) is probably of higher value.  

I like the direction of CMake, but this should be in a separate branch until it works.  I want to look around a bit as the things I've read tonight about CMake tend to agree with my (limited) experience with it.  I.e. not that good.  I really like the overall idea of generating the make files for each platform.  That is a very good goal.  I am just not sure that CMake is the right tool yet.

Can you split this up?  I thought I could cherry pick the individual commits, but I am not sure how to do that with GitHub.

PR #23, opened 2016-01-02
Fix maybe-uninitialized warning from GCC

thatcadguy:


thatcadguy:
Forgot to add a description before. This turned up when compiling as a possible usage before initialization warning using GCC 5.3.

PR #25, opened 2016-01-13
Fixes from Clang's static analyzer

thatcadguy:
In examples\multithread_plc5_dhp.c: this block will never be true as tag
is set to NULL in the while loop beforehand.

In lib\ab\eip_dhp_pccc.c: the deleted block is an exact duplicate of the
previous set of lines. Encoded name is being copied into data twice and
data is incremented twice incorrectly (correct me on this if I'm wrong)

### DIFF
--- a/examples/multithread_plc5_dhp.c
+++ b/examples/multithread_plc5_dhp.c
@@ -311,11 +311,6 @@ void thread_func(void *data)
         sleep_ms(100+random_min_max(0,100));
     }
 
-    if(tag) {
-        fprintf(stderr,"Thread %d done but tag still not destroyed.\n",tid);
-        plc_tag_destroy(tag);
-    }
-
     fprintf(stderr,"Thread %d completed work.\n",tid);
 
     mark_done(&done[tid]);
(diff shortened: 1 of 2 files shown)


kyle-github:
Thanks!  Hmm, I'd need to look at these closely.  After a quick glance, they seem right.  I wonder if I checked in the wrong code for eip_dhp_pccc.c?  More later.

thatcadguy:
No problem. Like I said, correct me if I'm wrong on that eip_dhp_pccc.c
change. I stared at it for a good 5 minutes wondering what I was missing.
I'm pretty sure on the other one; basically tag is declared without
initialization, then it enters the while block, and at the end of the while
block it is set to NULL.

thatcadguy:
For reference, you can see the output yourself by installing clang and running `scan-build -k -v -V make` in the top level directory.

kyle-github:
Sorry for the delay.  Yeah, weird.  This does look like a copy-paste error.  Oddly, it is not destructive.  I.e. if the code worked before the duplicate, it will work after because the data pointer is reset between the two copies.

PR #53, opened 2017-05-30
Write logs also on managed Console (C#)

mesta1:
To write the log in a C# console application I need this line of code.


mesta1:
To print the log from C# the "Enable the Visual Studio hosting process" under Debug tab must be unchecked.

Issue #69, opened 2018-05-04
provides a pkg-config file


scontini76:
Many libraries provides a pkg-config (.pc) file that helps to detect the library and provides the correct compiler and linker parameters to the build system. It would be nice if libplctag follows this practice. 

see: https://www.freedesktop.org/wiki/Software/pkg-config/

If you agree I could help you with a pull request on this issue.

kyle-github:
Oops!  I missed this one. 

Yes, please!   I would love to have some idea how to support this.

scontini76:
Please have a look at my pull request #70

kyle-github:
Sorry for the delay.   This was merged.   Thank you!

Do you know of a way to take the CMake version information and use that to change the version in the source code?

kyle-github:
Thanks, again!

I have merged this and done another release.

scontini76:
Glad that you like the changes. I'm not very skilled with CMake, I mostly use automate and recently meson, but I can have a look to pass define to compiler in CMake or how to prepare a config.h file, I'll work on it if you want, fill an issue so I'll remember to work on it in my spare time!

kyle-github:
Thanks, I will take a look and see if I can figure it out.   It should be possible.

PR #70, opened 2018-05-07
prepare and install the pkgconfig file libplctag.pc

scontini76:
as suggested in issue #69

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ project (libplctag_project)
 # the project is version 1.5
 set (libplctag_VERSION_MAJOR 1)
 set (libplctag_VERSION_MINOR 5)
+set (libplctag_VERSION_PATCH 8)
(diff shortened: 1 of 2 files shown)


kyle-github:
Hi @scontini76,

Hopefully you have seen the discussion on the forum about the license.   Over the last six months or so it has become clear that the LGPL is starting to hurt possible users.   There has been a change in how software is released in the last couple of years with more and more software being single, complete, statically linked executables.  Unfortunately, the LGPL makes it difficult for this kind of software to be built without effectively become the GPL.

I have researched several different licenses and I feel that the MPL 2.0 has a good balance between allowing more current development processes while still protecting the code and encouraging users to give back useful changes and fixes (as you have so generously done!).

Would you consider relicensing your contributions under the MPL 2.0 as well as the LGPL?  I would like to dual license the library under both licenses.

Best,
Kyle

PR #77, opened 2018-08-24
fix implicit-fallthrough from GCC

scontini76:
add compiler directives to avoid implicit-fallthrough warning in the specific code where the behaviour is valid.


kyle-github:
I am going to bounce this one as I have solved this in a different way.   Can you please check the latest master branch and see if it is still warning for you?

Issue #79, opened 2018-09-04
Reading from modern PLCs (LGX) via PCCC


bmhgit:
Hi Kyle,

I've been trying to read a Tag from a modern PLC (1769-L32E CLGX), which is equipped with CIP via PCCC. Clarification: I want to read via legacy PCCC command, even though CIP is working fine. For testing purposes. 
It looks like the library does not use the UCMM when assembling a PCCC packet.
![wirepccc](https://user-images.githubusercontent.com/42959032/45015660-ae08f780-b022-11e8-9d2a-a7da2b0ed9a8.PNG)

Debug output is as following(debug level 2):
thread(0001) 2018-08-04 09:13:33.508 WARN check_read_status:325 PCCC command failed, response code: 5
ERROR: Unable to read the data! Got error code -28: PLCTAG_ERR_REMOTE_ERR
thread(0001) 2018-08-04 09:13:33.510 WARN lib_teardown:121 Tag 0x55dc46192b70 at index 1 was not destroyed!

I can see in wireshark that the PLC never returns a response packet, just the request.

I did specify the path=1,0 variable, because as far as I know you need routing to the PLC backplane for modern PLCs.

Is this intended behaviour (no PCCC support for modern PLCs) or what might be happening?
I'm happy to provide further assistance.

By the way, 64-Bit Windows works fine when compiling with CMake.

kyle-github:
Hi,

Sorry I seem to have missed this question before now!   I have never tried to force the library to read a Logix-class PLC using PCCC.   Can you provide your initial tag creation string?   And debugging output with debug level 3 or 4? 

The library does support reading PLC 5 data from a PLC 5 using PCCC encapsulated in CIP/EIP.   It also supports reading a PLC 5 on DH+ through an Ethernet DH+ bridge (such as a DHRIO module in a ControlLogix).   The latter might be triggered when you have a path.  I'll have to look more closely at that code.

I am trying to remember how to decode the packet.  It looks like it is sending a read (0x0f CMD with 0x68 FNC) of... N7:0?  

Have you turned on the PLC 5 mapping in the ControlLogix?   I have never used that mode, but I seem to remember that there is some configuration you need to do on the ControlLogix side.

kyle-github:
Oh, and thanks for the CMake tip!  I still have not had a chance to set up Windows 10 64-bit.  I need to talk with a friend at Microsoft to get them to buy me a cheap copy of Windows.

bmhh:
Sure!
[...]

PR #83, opened 2018-10-17
Add PCCC-Read support to LGX class controllers 

bmhgit:
Hi Kyle, as discussed in the corresponding issue, here is how I patched the library. Take a look at what you might do differently, I looked at eip_cip.c for reference, maybe there is a cleaner way to integrate the separation of UCMM and standard packages...

### DIFF
--- a/src/protocols/ab/cip.c
+++ b/src/protocols/ab/cip.c
@@ -279,6 +279,8 @@ int cip_encode_path(ab_tag_p tag, const char *path)
         tag->dhp_src  = 0;
         tag->dhp_dest = 0;
         tag->use_dhp_direct = 0;
+        tag->use_ucmm = 1;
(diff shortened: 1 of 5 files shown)


kyle-github:
As discussed in issue #79, I took a slightly different approach than this.  It seems to work and fits the existing structure of the library a little bit better (more toward future plans).  So, I will be rejecting the PR, but not because it was not good code!  

I will be doing more separation of the different types of access in the future.  One big change, if I can get it to work, will be to pull out native CSP (PCCC) protocol PLCs (or fakes like CSP against Logix) into a separate library/protocol type.   I will separate CIP from PCCC etc. as I clean up the library structure to eventually support ModBus too.  Once PCCC is separate, then I should be able to support this directly without all the workarounds that are currently in process of packing PCCC into CIP.   That will result in a much cleaner library and more compatibility.

Thanks again very much for all the effort to figure out how to make this work at all and to make the patch!

PR #91, opened 2019-02-10
Add support for Android

thatcadguy:
Android's C library doesn't have pthread_cancel(); replace with pthread_kill()

### DIFF
--- a/src/platform/linux/platform.c
+++ b/src/platform/linux/platform.c
@@ -604,7 +604,11 @@ void thread_stop(void)
 void thread_kill(thread_p t)
 {
     if(t) {
+#ifdef __ANDROID__
+        pthread_kill(t->p_thread, 0);
+#else
         pthread_cancel(t->p_thread);
+#endif /* __ANDROID__ */
     }
 }
 


kyle-github:
Ooh, nice!

I am not sure I even use any thread killing calls.   I will have to check.   They are fraught with problems on various platforms (clean up is easy to skip).

kyle-github:
That was really all there was to get Android compilation to work?

thatcadguy:
Not 100% on it. With Termux, yes. My understanding is that Android is a modified form of Linux, and Termux is just a terminal emulator like bash, but with Debian packages. The Clang package that comes with Termux uses the Android NDK libs. You can build APKs with Termux so I'm assuming it's good enough.

kyle-github:
Hi @thatcadguy,

Would you be willing to release this patch under the MPL 2.0 instead of the LGPL of the original library?   Hopefully you have seen the discussion on the forum, but it is becoming more and more of a problem for users to have a license that makes it very hard to do a single stand-alone statically linked executable.   Unfortunately, the LGPL does make that difficult.   

I would prefer something a little "stronger" than the MPL 2.0 but the license seems to have the right combination of copy-left and protections in general.

Best,
Kyle

jake-edgenaut:
Hi Kyle, sure, no problem. Do I need to do anything? Best, Jake

PR #92, opened 2019-02-10
Prevent uninitialized variable reference in stress_api_lock.c

thatcadguy:
if rc != PLCTAG_STATUS_OK (line 94), then the variable 'value' will be uninitialized and is still referenced by the fprintf call on line 117. Clang spit up about this when I was compiling on Termux.

### DIFF
--- a/src/examples/stress_api_lock.c
+++ b/src/examples/stress_api_lock.c
@@ -81,7 +81,7 @@ void *test_tag(void *data)
 
     while(!done) {
         int rc = PLCTAG_STATUS_OK;
-        int32_t value;
+        int32_t value = 0;
         int64_t start;
         int64_t end;
 


thatcadguy:
Nevermind, I'm retarded. Just saw the incrementing of value. Need sleep.

kyle-github:
The question is why clang was saying it was not initialized.   I will check the code.  Usually a warning does indicate something is wrong!

thatcadguy:
The very first time it goes through the loop there's the potential to use value uninitialized if rc != PLCTAG_STATUS_OK on line 94, but you're declaring value within the loop. I forget what the rules are, very rusty on my C, but I believe if you declare/define value as static it only gets initialized once, which is what you want here.

thatcadguy:
```
/home/jake/libplctag/src/examples/stress_api_lock.c:94:12: warning: variable 'value' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]
        if(rc != PLCTAG_STATUS_OK) {
           ^~~~~~~~~~~~~~~~~~~~~~
/home/jake/libplctag/src/examples/stress_api_lock.c:117:109: note: uninitialized use occurs here
        fprintf(stderr,"Test %d, iteration %d, got result %d with return code %s in %dms\n",tid, iteration, value, plc_tag_decode_error(rc), (int)(end-start));
                                                                                                            ^~~~~
/home/jake/libplctag/src/examples/stress_api_lock.c:94:9: note: remove the 'if' if its condition is always false
        if(rc != PLCTAG_STATUS_OK) {
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jake/libplctag/src/examples/stress_api_lock.c:84:22: note: initialize the variable 'value' to silence this warning
        int32_t value;
                     ^
                      = 0
2 warnings generated.
```

kyle-github:
I think I have fixed these.   I wonder why I was not seeing warnings about this.   What compiler flags are you using?   

Please check the version that I just pushed to master.

thatcadguy:
Warning is gone! I didn't specify any additional flags for Clang, it's just the default stuff in CMakeLists.txt

kyle-github:
That is... odd.  I am not getting any warnings at all for that file :-(   Maybe I am not using clang?  I thought I set my default to clang.  I will have to check both GCC and clang.

thatcadguy:
Yes, odd. I'm using Clang 3.8.1. I invoke CMake with 'CC=clang CXX=clang++
cmake'.

kyle-github:
Hi @thatcadguy ,

I think I have a few other patches that you've send in over the past few years.   At least the repo owner seems to have the same name but the account is gone.  Was that you?

If so, would you consider relicensing your contributions under a dual license of MPL 2.0 and LGPL?   It is becoming clear that the new (what was old is new again) development process of creating stand-alone, statically linked executables is much more difficult with LGPL code.   The MPL 2.0 seems to offer most of the same protections and rights.

Best,
Kyle

jake-edgenaut:
Hi Kyle, yes that was me, I deleted my account and then came back with the
same username. You can relicense all of my contributions as you see fit, I
assign all my IP rights for those contributions to you. Best, Jake

kyle-github:
Thanks!

PR #103, opened 2019-07-17
src/wrappers: Add a simple wrapper for Golang

alimon:
- Export functions provided by libplctag.h under plctag.Func.
- Export defines for errors as constants.
- Add toogle_bool.go example to test the wrapper.

Signed-off-by: Aníbal Limón <limon.anibal@gmail.com>

### DIFF
--- a/src/wrappers/go/README
+++ b/src/wrappers/go/README
@@ -1,0 +1,10 @@
+A simple Go wrapper for libplctag
+
+1. Build and install libplctag.
+2. Set GOPATH to take into account this directory.
+
+$ export GOPATH=$(pwd):$GOPATH
+
+3. Run examples in go
+
+$ go run examples/toogle_bool.go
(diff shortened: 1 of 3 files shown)


kyle-github:
Thanks!   This looks great.   Thanks for putting in the license in the files.

kyle-github:
Hi @alimon, I want to change the license of the core library to MPL 2.0.   I will be splitting out the wrappers into separate repos so that they can more easily be found and maintained.   I would like to keep the core library and the wrappers with the same library, but it is not required.

Please let me know if you would agree to a license change.  Again, this is your code, so if you do not want to, it is perfectly fine!

Would you be willing to be the team (or at least on the team if you do not have time) to maintain the Go wrapper?   I have moved libplctag to its own GitHub organization to allow users a single place to find everything they need.   Each repo in the organization can have its own team to own and maintain it.   

Please let me know if either of these things are something you would consider.

Best,
Kyle

PR #106, opened 2019-08-13
improved golang example

garciaolais:


kyle-github:
Thanks!

kyle-github:
I will be splitting off the wrappers into separate repos in the near future.   I also would like to change the core library license to the MPL 2.0.   Would you have any objections to that?  The Go wrapper can still be maintained with LGPL if you prefer, but I am hoping to have all the wrappers and core use the same license to keep things simple.

Thanks for considering this.

Best,
Kyle

garciaolais:
I don't have any problem with that 🚀 👍

kyle-github:
Thanks!

TheFern2:
Does this wrapper work on win10?

PR #119, opened 2019-10-28
Initialize the socket fd to INVALID_SOCKET,

srdgame:
Initialize the socket fd to INVALID_SOCKET, as fd 0 is valid socket and may not create by plctag.

In our application, the fd 0 is created by libmosquitto.   If plctag cannot connect to device (e.g incorrect IP address), thus plctag closed the fd 0 which cause MQTT connection down.

### DIFF
--- a/src/platform/linux/platform.c
+++ b/src/platform/linux/platform.c
@@ -752,6 +752,9 @@ struct sock_t {
 
 
 #define MAX_IPS (8)
+#ifndef INVALID_SOCKET
+#define INVALID_SOCKET -1
+#endif
 
 extern int socket_create(sock_p *s)
 {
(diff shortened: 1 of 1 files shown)


kyle-github:
Thanks for finding this bug.  I filed bug #120 to track this.

I ended up solving this slightly differently.   I already had a flag in the socket struct to track whether or not the socket was completely open.   The Windows socket code looks like it was mostly handling this correctly.   I copied that over to the Linux platform.c and went through it all and fixed up a couple of places where the is_open flag was not being checked.

The fix I committed is a little bit more comprehensive and keeps the Windows and Linux code as close as possible.   Thus, I will not take this MR because I solved the problem in a different way.   

I want to keep as much platform independence as possible.  While I do not know of any platforms on which an fd could be -1, there might be one.   Since I had already created a special flag to track whether the socket was open or not, I could use that and not rely on special socket values.

I released version 2.0.27 to fix this bug.

Thanks again for finding this problem!

srdgame:
@kyle-github It is fine for me, as the bug closed.  Thanks, I have merge your code into my repo which has an lua binding.

kyle-github:
Please double check the changes I made.  They passed my testing, but I am probably not covering your use case correctly!

Any chance you could share the Lua bindings?   I am sure other users would be interested!

srdgame:
@kyle-github Will  try to create an pull request for Lua binding later.

PR #136, opened 2020-02-15
Fixes build on OS X.

mavenraven:
This works @kyle-github , may or may not be how you actually want to do it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,8 +52,6 @@ set ( util_SRC_PATH "${base_SRC_PATH}/util" )
 set ( example_SRC_PATH "${base_SRC_PATH}/examples" )
 set ( test_SRC_PATH "${base_SRC_PATH}/tests" )
 
-# OS-specific files for the platform code.
-# FIXME - does this work for macOS?
 if (UNIX)
     set ( platform_SRC_PATH "${base_SRC_PATH}/platform/linux" )
 elseif (WIN32)
--- a/src/platform/linux/platform.c
+++ b/src/platform/linux/platform.c
@@ -41,6 +41,35 @@
 #include <errno.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+/***************************************************************************
+ * On OSX, sys/socket.h is defined with:                                   *
+ * ...                                                                     *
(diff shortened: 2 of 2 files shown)


kyle-github:
Hmm.   I would rather not copy a define like that.   I am worried that this could be fragile if Apple changes the definition (admittedly unlikely as that would break old binaries).

I have another attempt based on some Stack Overflow items and other things I found online plus the header file snippet you sent.   Can you give the updated macos_fix branch a try again?

The core of this attempt is that the #ifdef is triggered if either _POSIX_C_SOURCE is _NOT_ defined or _DARWIN_C_SOURCE _IS_ defined.   So, if I am certain it is macOS (possibly iOS as well?), then I defined _DARWIN_C_SOURCE to be the same as _POSIX_C_SOURCE.   From what I understood from Stack Overflow, they are both date-like numbers (20080101 etc.).

mavenraven:
That compiles if you move 
```
#if defined(__APPLE__) || defined(__FreeBSD__) ||  defined(__NetBSD__) || defined(__OpenBSD__) || defined(__bsdi__) || defined(__DragonFly__)
    #define BSD_OS_TYPE
    #if defined(__APPLE__) && !defined(_DARWIN_C_SOURCE)
        #define _DARWIN_C_SOURCE _POSIX_C_SOURCE
    #endif
#endif
```
before the relevant include.

I would suggest making it this instead:
```
#if defined(__APPLE__) || defined(__FreeBSD__) ||  defined(__NetBSD__) || defined(__OpenBSD__) || defined(__bsdi__) || defined(__DragonFly__)
    #define BSD_OS_TYPE
    #if defined(__APPLE__) && !defined(_DARWIN_C_SOURCE)
        #define _DARWIN_C_SOURCE
    #endif
#endif
```

kyle-github:
I have an alternate solution.   I was able to add a macOS build to the Azure pipeline and used this to figure out what to change in the CMake file to make the build work "natively."   I also renamed the "linux" platform code to "posix" which is not strictly accurate since most of the SIGPIPE avoidance code is not POSIX compatible :-/

mavenraven:
Ah nice, wasn't aware the azure pipelines had Mac build machines.

PR #152, opened 2020-06-29
OMRON NJ/NX Series Ethernet/IP initial support.

j2mcv:
Proposed changes to add support for OMRON NJ/NX PLCs.

- Added a new cpu (AB_PLC_OMRON_NJNX) under the ab_eip protocol.

- Uses the existing code paths for MLGX800, I've added a new tag_vtable_t to handle cases where cpu doesn't support partial data and extra parameters to the build_[read|write]_request functions.

- I also had to set to zero the ip address length in match_ip_addr_segment to make in work in Windows under MSVC.

### DIFF
--- a/src/protocols/ab/ab_common.c
+++ b/src/protocols/ab/ab_common.c
@@ -266,6 +266,11 @@ plc_tag_p ab_tag_create(attr attribs)
         tag->allow_packing = 0;
         break;
 
+    case AB_PLC_OMRON_NJNX:
+        tag->use_connected_msg = 1;
+        tag->allow_packing = 0;
+        break;
+
     default:
         pdebug(DEBUG_WARN, "Unknown PLC type!");
         tag->status = PLCTAG_ERR_BAD_CONFIG;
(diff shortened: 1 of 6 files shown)


kyle-github:
Wow!  Nice work.   I was actually thinking of just copying the entire Logix code set and then removing all the unconnected support.   This would result in more duplication at first, but so far I have found that each PLC type seems to have quirks of its own and eventually I want them separate.  I just had to do that with the PCCC-PLCs like PLC/5 and SLC.

There is a lot of similarity between the Micro800 and the OMRON.   Does the OMRON allow you to list tags?   The Micro800 does not.   Now that I am thinking about this more, it might make sense to split out the Micro800 and OMRON into a separate code set from the ControlLogix and CompactLogix.

I notice that you are trying to merge this against the "add_modbus" branch.   That branch is likely to stay separate for a while as the Modbus code is not at all ready for use.   Can you please rebase against the "add_omron" branch?   I am probably not going to get much of a chance to look deeply at splitting up the existing code until this coming weekend.

j2mcv:
> Wow! Nice work. I was actually thinking of just copying the entire Logix code set and then removing all the unconnected support. This would result in more duplication at first, but so far I have found that each PLC type seems to have quirks of its own and eventually I want them separate. I just had to do that with the PCCC-PLCs like PLC/5 and SLC.
> 

Thanks! I was trying precisely to avoid more duplications

> There is a lot of similarity between the Micro800 and the OMRON. Does the OMRON allow you to list tags? The Micro800 does not. Now that I am thinking about this more, it might make sense to split out the Micro800 and OMRON into a separate code set from the ControlLogix and CompactLogix.
> 

I think it makes sense.

I've tested the list tags feature without success, seems something very specific of of these models. I will investigate if the OMROM has something similar.

> I notice that you are trying to merge this against the "add_modbus" branch. That branch is likely to stay separate for a while as the Modbus code is not at all ready for use. Can you please rebase against the "add_omron" branch? I am probably not going to get much of a chance to look deeply at splitting up the existing code until this coming weekend.

I'll do.

Issue #157, opened 2020-07-16
Question: how to read the tag from Rockwelll Micro 850 (2080-LC50-24QWB)?


abonent312:
Hi! Pls help me to build the correct path-string for Rockwelll Micro 850 (2080-LC50-24QWB)?
I used the next form (in src/examples/simple.c):
`protocol=ab-eip&gateway=192.168.5.20&path=1,0&cpu=micro800&elem_count=1&elem_size=4&debug=4&name=L19_0_8_DispActWeight[1]`
it gives error:

`2020-06-16 21:26:48.097 thread(2) tag(0) WARN recv_forward_open_resp:2193 Forward Open command failed, response code: Port Not Available (1)
2020-06-16 21:26:48.097 thread(2) tag(0) WARN recv_forward_open_resp:2214 CIP error code 1!
2020-06-16 21:26:48.097 thread(2) tag(0) INFO recv_forward_open_resp:2234 Done.`

(the 
[libplctag_1_0.txt](https://github.com/libplctag/libplctag/files/4934549/libplctag_1_0.txt)
full log is attached)

I tried different combinations for "path=x,y" : 2,0 3,0 4,0. It helps a bit, session is created
`2020-06-16 21:25:58.708 thread(3) tag(0) DETAIL session_create_request:2384 Done.
`

but after that I see another error (TIMEOUT):
`2020-06-16 21:25:58.710 thread(3) tag(0) DETAIL send_eip_request:1728 Done.
2020-06-16 21:25:58.744 thread(3) tag(0) INFO send_forward_open_req_ex:2156 Done
2020-06-16 21:25:58.744 thread(3) tag(0) INFO recv_forward_open_resp:2169 Starting
2020-06-16 21:25:58.745 thread(3) tag(0) DETAIL recv_eip_response:1748 Starting.
2020-06-16 21:26:03.685 thread(1) tag(0) WARN plc_tag_create:560 Timeout waiting for tag to be ready!
2020-06-16 21:26:03.686 thread(1) tag(0) DETAIL ab_tag_abort:685 Starting.
2020-06-16 21:26:03.686 thread(1) tag(0) DETAIL ab_tag_abort:701 Done.
2020-06-16 21:26:03.686 thread(1) tag(0) WARN plc_tag_create:567 Error PLCTAG_ERR_TIMEOUT while trying to create tag!
`
thank you!

kyle-github:
Hi,

[...]

PR #159, opened 2020-07-18
Fix micro800 regression

kyle-github:
Run all tests on both x86 and x64 Ubuntu.


kyle-github:
Fix #157, regression in Micro800.   Expand CI tests to include Micro800.

PR #171, opened 2020-07-29
Limit elem_count to 1 on OMRON PLCs.

j2mcv:
OMRON PLCs return the full size of an array, so we don't allow elem_count != 1.

### DIFF
--- a/src/protocols/ab/ab_common.c
+++ b/src/protocols/ab/ab_common.c
@@ -368,7 +368,6 @@ plc_tag_p ab_tag_create(attr attribs)
         pdebug(DEBUG_WARN, "Unknown PLC type!");
         tag->status = PLCTAG_ERR_BAD_CONFIG;
         return (plc_tag_p)tag;
-        break;
     }
 
     /* pass the connection requirement since it may be overridden above. */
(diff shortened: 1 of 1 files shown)


kyle-github:
Thanks!   

I think it might be time to pull the Omron stuff out into its own set of code, even though that will make a lot of duplication.   It is a weird PLC, isn't it?

I will try to refactor the Omron code out this weekend and see if can cut down on the duplicate and complicated code paths.   For instance, I suspect that the Omron cannot handle unconnected messaging (UCMM).   That is a big part of eip_cip.c.

What happens if you do send an element count of something other than 1?

src/protocols/ab/ab_common.c
```diff
@@ -380,10 +379,26 @@ plc_tag_p ab_tag_create(attr attribs)
...
+
+    switch(tag->plc_type) {
+    case AB_PLC_OMRON_NJNX:
+        if (tag->elem_count != 1) {
+            tag->elem_count = 1;
+            pdebug(DEBUG_WARN,"Attribute elem_count should be 1!");
+        }
+
+        /* from here is the same as a AB_PLC_MLGX800. */
+
```
kyle-github, line 391: So this falls through?   I think the compiler will complain about that.  I will need to check.  Generally I try to have fall through only on cases where each `case` branch is identical.   Otherwise I try to break out each `case` branch as completely independent code.

kyle-github:
Note that this has not been added to the prerelease for 2.1.1 yet.   It is just at the tip of the `prerelease` branch.

Thanks for these fixes!

src/protocols/ab/ab_common.c
j2mcv: > Thanks!
> 
> I think it might be time to pull the Omron stuff out into its own set of code, even though that will make a lot of duplication. It is a weird PLC, isn't it?
> 
> I will try to refactor the Omron code out this weekend and see if can cut down on the duplicate and complicated code paths. For instance, I suspect that the Omron cannot handle unconnected messaging (UCMM). That is a big part of eip_cip.c.
> 
> What happens if you do send an element count of something other than 1?

A request error, don't have access to the error code ATM.

I agree that it will be better to give OMROM his own code.

kyle-github:
Hi Jorge,

I have had to change the Omron code.   Unfortunately, the way that it worked broke handling for tags larger than the packet size.   It broke the handling in such a way that the library hangs when reading a large tag :-(

I have tried to set up Omron support by forcing the read/write commands to non-fragmenting when the PLC type is Omron.   This seems to work with my hacked emulator program, ab_server.   The packets all get the right CIP commands for the right types of PLCs.   The byte offset is not put into the packet when I select Omron as the PLC type.

I just released version 2.1.13 as a pre-release.  Can you please check it to make sure I did not break the Omron functionality!

j2mcv:
Hi Kyle,

unfortunately I will not be able to test it until Aug 24th.

Sorry for the inconvenience.

kyle-github:
No problem, I am going to be out for a while as well.   I hope you are getting a vacation!

j2mcv:
Yes, thank you!

kyle-github:
I released 2.1.14 as a pre-release.   This has some additional code, but the Omron changes are all there as well.  I have Omron running in the CI tests on Linux and macOS now.

j2mcv:
I've tested the prerelease version (2.1.15) on a Nx1P2 and works as expected.

kyle-github:
Thanks, @musiguello !

PR #219, opened 2020-12-09
set auto_sync_last_read as time_ms()

einerfreiheit:
When reading takes 1 ms or more  _we are pushing the next read out_.

It seems that the read time even on the local network can be more than 1ms which means the actual sync period is 2 * auto_sync_read_ms.

I suggest using the time at the end of the reading.
**Update:**
We check the condition `if(tag->auto_sync_last_read + tag->auto_sync_read_ms <= time_ms())` to start reading, that means _we are_ always _behind_ even if the reading time is 0.

### DIFF
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -299,6 +299,7 @@ THREAD_FUNC(tag_tickler_func)
 
                             /* if we have automatic read enabled, make sure we set up correct times. */
                             if(tag->auto_sync_read_ms > 0) {
+                                /* calculate elapsed_sync_peroids instead using current time to avoid read times drift*/
                                 int64_t elapsed_sync_peroids = (time_ms() - tag->auto_sync_last_read) / tag->auto_sync_read_ms;
                                 tag->auto_sync_last_read += elapsed_sync_peroids * tag->auto_sync_read_ms;
                             }
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -299,13 +299,7 @@ THREAD_FUNC(tag_tickler_func)
 
                             /* if we have automatic read enabled, make sure we set up correct times. */
                             if(tag->auto_sync_read_ms > 0) {
-                                /* when do we read again? */
-                                tag->auto_sync_last_read += tag->auto_sync_read_ms;
-
-                                /* if we are behind, catch up by pushing the next read out. */
-                                if(tag->auto_sync_last_read <= time_ms()) {
-                                    tag->auto_sync_last_read = time_ms() + tag->auto_sync_read_ms;
-                                }
+                                tag->auto_sync_last_read = time_ms();
                             }
 
                             events[PLCTAG_EVENT_READ_COMPLETED] = 1;
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -299,7 +299,8 @@ THREAD_FUNC(tag_tickler_func)
 
                             /* if we have automatic read enabled, make sure we set up correct times. */
                             if(tag->auto_sync_read_ms > 0) {
-                                tag->auto_sync_last_read = time_ms();
+                                int64_t elapsed_sync_peroids = (time_ms() - tag->auto_sync_last_read) / tag->auto_sync_read_ms;
+                                tag->auto_sync_last_read += elapsed_sync_peroids * tag->auto_sync_read_ms;
                             }
 
                             events[PLCTAG_EVENT_READ_COMPLETED] = 1;


kyle-github:
Hi @einerfreiheit¸thanks for proposing this.   I think neither logic is correct, but that is subject to interpretation.

What I wanted to do here is make sure that the read times do not drift.   Simply assigning the last read time to the current time causes drift.  By drift, I mean that if we are not careful, the processing time of the read code starts to change the read time:

- read at 100ms
- processing takes 5ms.
- next read at 205ms.
- processing takes 5ms.
- next read at 310ms.
...

What I want is consistent times:

- read at 100ms
- processing takes 5ms, so next read in 95ms.
- read at 200ms
- processing takes 5ms, so next read in 95ms.
- read at 300ms.

If you are trying to get clean timings on your samples it is very easy to generate drift.

I will look into this more carefully later today (I need to deal with $day_job first) and see what to do here.

The response time is not going to be exact.   There are too many variables controlled by the OS itself (total number of threads in the systems, general load, scheduling time etc.).   So that logic was to try to keep the drift from impacting the read times overall.  There will be some jitter, but I want to make sure it does not grow.

einerfreiheit:
Hi @kyle-github, thank you for replay!
It seems that we still need to know how long does it take to get response. 
My next suggestion is:

- calculate  elapsed time since last sync (`time_ms() - auto_sync_last_read`)
- calculate  number_of_periods  to add: (`(int) (elapsed_time / auto_sync_read_ms)`)
- update auto_sync_last_read: `auto_sync_last_read += number_of_periods * auto_sync_read_ms` 

Case 1 (read every 100ms, processing takes 5ms):

- auto_sync_last_read  = 0, read at 100ms, processed at 105ms, number_of_periods  = 1, set auto_sync_last_read at 100ms
- auto_sync_last_read  = 100, read at 200ms, processed at 205ms, number_of_periods  = 1, set auto_sync_last_read at 200ms
...

Case 2 (read every 100ms, processing takes 105 ms):
- auto_sync_last_read  = 0, read at 100ms, processed at 205ms, number_of_periods  = 2, set auto_sync_last_read at 200ms
- auto_sync_last_read  = 200, read at 300ms, processed at 405ms, number_of_periods  = 2, set auto_sync_last_read at 400ms
...
In this way we skip next sync but avoid time drifting.

Or maybe your do it your way :)

kyle-github:
There are a few cases we need to cover:

1. We are starting up and `auto_sync_last_read` is zero.
2. We have had a normal run and the jitter is less than `auto_sync_read_ms`.
3. We had a slow run and the jitter is more than `auto_sync_read_ms`.

The first and last cases are similar.   We have two choices for those: a) we add some multiple of `auto_sync_read_ms` to get the next read time later than the current time, b) we reset the next read time to the current time plus some offset (perhaps `auto_sync_read_ms`).

I was doing method b above.   Your proposal is method a.   I like that better.  I think the code would be something more like:

```c
if(tag->auto_sync_read_ms > 0) {
    int64_t periods = 1;

    /* Round up to the next period. This tries to eliminate any jitter from accumulating. */
    periods = (time_ms() - tag->auto_sync_last_read + (tag->auto_sync_read_ms - 1)) / tag->auto_sync_read_ms;

    tag->auto_sync_last_read += periods * tag->auto_sync_read_ms;

    if(periods > 1) {
        pdebug(DEBUG_WARN, "Skipping " PRid " periods to catch up on lost samples!", periods -1);
    }
}
```

I will need to look at some cases to make sure that it works correctly.  

Alternatively, I could just calculate the total periods since time 0 and keep the auto sync tags all aligned with the clock.   That sounds good but then I am worried about too many things lining up exactly and hitting the host CPU and PLC fairly hard in bursts.   Of course, that would help when you have a *Logix and can multiplex request packets.   So perhaps that is better?  Hmm... then you could not print the warning as easily.

Thanks for bringing this up!  I think this needs a little more checking and thought before I commit to a fix.

kyle-github:
Hmm, no, I think you were right.  You probably do not want to round up.   I think there is an edge case there but I need to play with numbers to make sure that it is right.

kyle-github:
I ended up doing this slightly differently.   

I changed how automatic reads are triggered by tracking the next read and changing where the scheduling was done.   There are a couple other changes I made in commit https://github.com/libplctag/libplctag/commit/655da3361a52b69c380b84959e21b3fcefbf6436.

Please check through that commit to see if that fixes the problem you found!

einerfreiheit:
Yep it works, thanks!

kyle-github:
Thanks for finding this!   I hope to push out the current prerelease branch to release later this week.   This fix will be in that release, thanks to you!

PR #232, opened 2021-02-06
Fix logging callback debug level

busyboredom:
The logging callback is currently passed the global debug level (or verbosity). Is this intentional? 

It would be much more useful if the actual debug level of the current message were passed instead.


kyle-github:
Thanks!   This does look like a miss on my part.  The API docs do not state one way or another what is returned, but your patch makes more sense to me than what I was doing!

kyle-github:
I updated the API docs to match what your patch does.   Thanks again!

Issue #162, opened 2020-07-22
API addition: Support strings more cleanly across all PLC types


kyle-github:
Support strings as a "native" type of the library, in the API.  For example:

```c
const char *plc_tag_get_string(int32_t tag_id, int offset);
int plc_tag_set_string(int32_t tag_id, int offset, const char *string_val);
```

As with other calls, the `offset` parameter is in bytes.  For the getter, return NULL on failure either due to lack of data, an out of bounds offset or other error.

Note that this will require some sort of internal byte swapping and other handling to be done as some PLCs return raw string data in various forms and orders.   

Challenges:  

1. PLCs that return only the valid string bytes may not have a simple way to determine the maximum string size.   At least with AB PLCs, strings are 82 characters _except_ on Micro800 where they appear to be 255 characters.  There are also packet size limitations that may impact this.
2. How to handle arrays of strings?   PLCs that allow writes to arrays of strings may require significant changes to the representation in memory before writing the values to the PLC.  PLC/5 and Control/CompactLogix appear to transmit the full 82 characters in all cases.
3. This is a core library feature that would interpret values that were not scalar values or arrays.   This breaks the existing pattern of the library performing just the core functionality and it thus arguably not appropriate for the core library.

kyle-github:
How would the offset be used in the case of string arrays?   Set elem_size to let the application know how big a fixed size string is?

While low friction for apps, this may be too high level for the core library.

kyle-github:
Another possible variant of this would be to rewire the existing getters and setters to handle the byte swapping etc. automatically.   This gets challenging when a string is a field within a UDT tag.   That alone argues for new API calls.

kyle-github:
Perhaps a different, slightly lower level set of API calls:

```c
int plc_tag_get_string_length(int32_t tag_id, int string_start_offset);
int plc_tag_set_string_length(int32_t tag_id, int string_start_offset, int string_len);
int plc_tag_get_string_char(int32_t tag_id, int string_start_offset, int char_index);
int plc_tag_set_string_char(int32_t tag_id, int string_start_offset, int char_index, int char_val);
```

This is a little clunky, but does not require any additional memory allocation or in place data transformation.  The tag buffer would still contain unchanged raw PLC data.  This would allow configuration of the size of the count element, if there was one so that it can be translated in get/set string size.  It would allow handling for byte-swapping of the string character indexes.

It abstracts the underlying byte representation sufficiently that generic code can be written to work with any PLC type at the application level.

[...]

PR #234, opened 2021-02-08
New access methods for bulk read/write operations

fluisgirardi:
New access methods for bulk read/write operations and updated Pascal Wrapper to these new functions. 
Plus: string access functions are now in the Pascal wrapper (you warned me and I didn't understand)


kyle-github:
Thanks for this!   Can you please base the PR on the `prerelease` branch?   I keep the `release` branch as the last fully tested, verified release and there are a few things (see below) that may need to be changed.  For small fixes that do not impact the core library I will sometimes accept PRs directly onto the `release` branch.

There was some discussion in issue #162 where [I proposed a different API](https://github.com/libplctag/libplctag/issues/162#issuecomment-720109046).   It ends up being effectively the same.

Before I can including this I (or you) will need to add:

- API documentation on the [API wiki](https://github.com/libplctag/libplctag/wiki/API).
- An example program showing the use.
- A fix for the version number.   I am trying to enforce SemVer so this would be 2.3.0.

I am working on code to support raw CIP commands and several other things related to tag listing as part of the new event core, so I think the example program can wait.  

I may rename the functions slightly as "block" does not really indicate what kind of data is being returned.  I think having the word "byte" or "raw bytes" in the name would make it very clear what was being returned.   Perhaps something like this?

```c
int plc_tag_set_raw_bytes(int32_t id, int offset, uint8_t *bytes, int byte_count);
int plc_tag_get_raw_bytes(int32_t id, int offset, uint8_t *bytes, int byte_count);
```
I will work on the API docs and the version number and get that into this PR first before I merge it into the release branch. 

Also, are you OK with licensing these additions under the MPL 2.0 and LGPL licenses?  The entire C library is under those licenses and I want to keep the licensing simple for users.  

Thanks again for this great contribution!

fluisgirardi:
Of course. Give me some time to do that.  I'm with work in progress with
these changes.

Em seg, 8 de fev de 2021 22:49, Kyle Hayes <notifications@github.com>
escreveu:

kyle-github:
I updated the [API docs for the new raw data access functions](https://github.com/libplctag/libplctag/wiki/API#handling-raw-bytes).  Please take a look and send me any corrections.

kyle-github:
I found a way to change the base branch!  I am merging this to the prerelease branch first and then I'll fix up the names.

PR #235, opened 2021-02-11
Updated Python Wrapper

GitHubDragonFly:
- Updated code for C library loading
- Folder structure to be used can be seen in the included pictures (Multi OS use)
- Optionally, only the corresponding OS folder/library can be created (Single OS use)
- Added String data accessor functions, which do require release 2.2.0 libraries at minimum


kyle-github:
Thanks!

Issue #218, opened 2020-11-29
Add ARM as a CI target


timyhac:
As discussed on https://github.com/libplctag/libplctag.NET/issues/111, using libplctag on ARM devices is becoming increasingly popular. This issue requests that runtimes are produced for ARM.

Github Actions does not have free hosted ARM runners, but cross-compiling might be possible.

kyle-github:
Yep.   Now a high priority.   I will start w/Raspian/Debian since I think I can build that under Ubuntu.   After that, I'll try to see if I can build ARM and ARM64 under Windows.   

First, solve the CPU load problem...

timyhac:
I don't have the full answer yet, but I have been able to cross-compile for ARM 32bit Ubuntu running on Raspberry Pi 3. The compilation was done on x64bit Ubuntu running in a Virtual Machine on my Windows 10 laptop.


I don't know at this stage how this would fit into CMakeLists.txt, so I just plonked the following on line 80 (after all other commands to configure the build tools).


```
set(CMAKE_SYSTEM_PROCESSOR arm)
set(CMAKE_C_COMPILER arm-linux-gnueabihf-gcc)
set(CMAKE_CXX_COMPILER arm-linux-gnueabihf-g++)
```

The relevant build tools need to be installed:
```sudo apt-get install g++-arm-linux-gnueabihf```


Seems reasonably straightforward, but I'm not across CMake and how this would fit into CI.


Relevant information:
* https://askubuntu.com/questions/250696/how-to-cross-compile-for-arm
* https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-linux


Update: also works on Raspbian.

**Comment by kyle-github** [comment]
I found another slightly higher-level way to do it and added that into the CI process.   It should be building for Arm v6hf (hard float), Arm 7l (no idea what the "L" is for, little endian?), and Aarch64 now.  I also started building for Windows for Arm 32-bit and Arm 64-bit.

Instead of including the compiler itself, Debian and Ubuntu have another package that pulls in all the basics:

```text
sudo apt install crossbuild-essential-armhf crossbuild-essential-arm64
```
That gets the compiler and all the base libraries and include files.   The C libraries builds like a charm with that.   

[...]

PR #243, opened 2021-02-18
Add Windows ARM builds to CI pipeline

timyhac:
Relates to https://github.com/libplctag/libplctag/issues/218

I've tried the ARM build and it does work, but haven't tried the ARM64 build.

Of course the tests won't work because the runner itself is x86 or x64 and the binaries are for ARM/ARM64, so they've not been added to the steps.

Relevant documentation in cmake: https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2016%202019.html

I don't know what the purpose of the ci.yml.in file is, but I do note that the build fails without it.
An empty file with the right filename seems to get the build working, but I can reinstate the file if preferred.


timyhac:
I think I've done this the wrong way, it should be a branch on the forked repository, and the Pull Request into release on the primary repository.

kyle-github:
Er, yeah, the ci.yml.in is the base, generic input file for the GitHub Actions CI process.   The CMake file takes that, updates things like the current version number (which is control exclusively in CMake) and builds the ci.yml file as a product of the CMake run itself.

So, ci.yml.in is the root file.   I have to check in the ci.yml because I have not figured out how to make the CI process generate its own YAML control on the fly.  So ci.yml is really a generated file, but because I am a lazy ass I just check it in after updating CMake and ci.yml.in.

The way to edit this is to edit ci.yml.in, copy one of the Windows sections, add in the right build arguments, and remove all the test sections.   I will take a peek at this later tonight now that you have done all the hard work.   There are instructions for how to do this by hand in the BUILD.md file.

PR #249, opened 2021-03-01
replace the special (non minus) char with a minus - this makes string output nice on windows

Laro88:
It seems be a special unicode line that looks like minus. Windows build with multibyte dislike it :-)

### DIFF
--- a/src/examples/list_tags.c
+++ b/src/examples/list_tags.c
@@ -556,16 +556,16 @@ void print_element_type(uint16_t element_type)
 
         switch(atomic_type) {
             case 0xC1: type = "BOOL: Boolean value"; break;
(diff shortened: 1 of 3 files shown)


kyle-github:
Wow.   Thanks!   I wonder how long that has been there?  I will need to make a pass to make sure that no other multi-byte characters are lurking in the source!

Again, thanks for this!

Laro88:
He he - I was really puzzled why I got this odd printout, thought for a splitsecond that printf has a problem with [numerical] succeded by "-" causing a to me unknown math operation to happen - and then I looked a bit closer in the memory for the strings :-)

I did not find any more of those "silly minuses" but I only checked this branch.

PR #267, opened 2021-06-13
add support for a basic CLI interface

atifali:
This adds a standalone CLI program to the core library. 

Refer to the [README](https://github.com/atifali/libplctag/blob/plc-cli/src/cli/README.md) for details.


atifali:
@kyle-github this is a super handy addition that we have been using for a while. Hope this could be useful to other people.

JSON format line output is particularly useful when being consumed by external programs.

kyle-github:
Hi Atif (Atif Ali?).   Thanks for this contribution!   I am traveling for work for the next couple of weeks but hopefully I will have time to get this merged.   

Can you please include a license for the code?  I see `uthash.h` has its own license.

I will merge this into src/contrib since this is a complete tool by itself and at least `uthash.h` is under a separate license.  I will try to get this done before I leave again.

Thanks again for this great contribution!

CMakeLists.txt
```diff
@@ -140,6 +140,10 @@ if (${CMAKE_GENERATOR} MATCHES "MinGW Makefiles")
     SET(CMAKE_CXX_FLAGS "-m32 -mno-ms-bitfields -D_WIN32_WINNT=0x0600 -DLIBPLCTAGDLL_EXPORTS=1")
 endif()
 
+# set flags for MacOSX
```
kyle-github, line 143: There is an Apple section in the compiler part.  Can this be moved there?

CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ set (VERSION "${libplctag_VERSION_MAJOR}.${libplctag_VERSION_MINOR}.${libplctag_
 
 set (LIB_NAME_SUFFIX "${libplctag_VERSION_MAJOR}.${libplctag_VERSION_MINOR}.${libplctag_VERSION_PATCH}")
 
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 2.8...3.19)
```
kyle-github, line 41: I am not sure about this one.   Why do you want to limit the version of CMake to <= 3.19?   Is there a breaking change at that point?

CMakeLists.txt
```diff
@@ -255,6 +260,7 @@ elseif (CMAKE_C_COMPILER_ID STREQUAL "MSVC")
     # using Visual Studio C/C++
     set(BASE_RELEASE_FLAGS "${CMAKE_C_FLAGS} /DLIBPLCTAGDLL_EXPORTS=1 /W3")
     set(BASE_DEBUG_FLAGS "${CMAKE_C_FLAGS} /DLIBPLCTAGDLL_EXPORTS=1 /W3")
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
```
kyle-github, line 263: IIRC, this will cause the compiler to not throw warnings about some older standard C functions.  I have avoided that in the core library by using the Windows versions (blah_r or blah_s usually) of these functions.   I would prefer to keep the warnings on.   Microsoft does change what things warn and what things just do not compile from version to version, so I think this may be a dangerous change.   Can you try to make it compile on Windows without this?  Fine if you cannot, but it would be nice if it could.

CMakeLists.txt
```diff
@@ -529,6 +535,34 @@ else()
     CONFIGURE_FILE("${CMAKE_CURRENT_SOURCE_DIR}/src/lib/libplctag.h" "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libplctag.h" COPYONLY)
 endif(ANDROID_BUILD)
 
+# add the cli
```
kyle-github, line 538: Hmm, now that I am really looking into this.  I wonder if it would make more sense to break the entire CLI code out and make it have its own CMake file?  Then you could have a much simpler build process.

src/cli/README.md
```diff
@@ -0,0 +1,60 @@
...
+                [-debug] [-interval] [-attributes] [-offline]
+
+        CLI Action (Required):
+
+        --read          - Perform a one-shot READ operation.
+        --write         - Perform a one-shot WRITE operation.
+        --watch         - Perform a continuous WATCH operation at the specified interval.
+        -h | --help     - Prints the Usage details.
+
+        LIBPLCTAG Parameters (Required):
```
kyle-github, line 28: Why do you want to break out the different tag params?   Different kinds of PLCs will use different parameters so this may be a bit of a problem in the future.   Is there something in your code that needs this separation?  In tag_rw and tag_rw2, I have a single parameter for every tag.   It is the same in the `data_dumper.c` example.  Each tag is a complete single string.

src/cli/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+        -protocol       - type of plc protocol. (default: ab_eip)
+        -ip             - network address for the host PLC. (default: 127.0.0.1)
+        -path           - routing path for the Tags. (default: 1,0)
```
kyle-github, line 32: Some PLC types explicitly warn if they have a path.  For instance all the PCCC-based PLCs.

kyle-github:
I have had a long delay on this due to legal requirements from my employer.  That is all sorted out now and I will take this up again.   I am going to need to move this to the contrib section as there is no license for this and, as noted above, uthash.h has a different license.

PR #270, opened 2021-07-03
static build for windows

Joylei:
this PR makes below changes:
- adds static build for windows
- new flag to opt out examples build


kyle-github:
Thanks for this!  I had to fix a few merge conflicts.  I will be doing some testing over the next couple of days and then should be able to push this as a release.

Issue #212, opened 2020-10-29
Passing some user data to the event callback


nikitapogodin:
When using the library from C++ code, it would be very useful to be able to pass user data when registering callbacks and return them back in callback calls.
This for example can be used to bind a callback to a C++ class instance (passing _this_ pointer as user data)

API could looks like:
void plc_tag_register_callback(int32_t tag_id, void (*tag_callback_func)(int32_t tag_id, int event, int status, void* used_data), void* user_data);

kyle-github:
Hi, thanks for proposing this.   I started with this exact API.  It is "traditional" for C-like languages.   Unfortunately, it does not work well for languages that move data around such as any language with generational GC.   I.e. Java, C#...   The problem is that the pointer becomes invalid the moment that GC runs and moves the object in memory.   There is no way to tell the C code that the stored pointer must be updated.

In C++, I suggest making a static map between the int32_t of the tag handle ID and the C++ object you want.   Then when the callback is called, you can look up the C++ object.  It is not quite as ergonomic as getting a pointer to a C++ object instance, but it does allow the same API to be used for all the wrapper languages.

This last point is what led me to the current API: it had to work for all possible wrapper languages.

If you know of a way to solve the GC problem, I would love to add an API for callbacks that allowed passing in data!  Really, if you have any idea at all how to solve this, I am very interested!

Please re-open if you have any ideas!

PR #345, opened 2022-01-26
callback with user data

Joylei:
This PR makes it possible to pass user data to the callback function, as #212 requested.

A new api was added
- plc_tag_register_callback2

It keeps compatible with existing code.

As to the concerns about GC, it's caller's responsibility to keep callbacks and user data on the stack, and free them after usage.

Let me know if you have any questions.


kyle-github:
Hi @Joylei, thanks for this contribution.  This is pretty much how I would have done it.   I want to hold on to this for a bit while I find out what is possible and what is not within Java and specifically JNA and find out from @jkoplo and @timyhac what works for C#.  API changes are set in stone once they happen so I want to make absolutely sure this is the right way to do it before I add this function.  I can always add API functions, but not change or remove them without breaking old code. 

I wonder if rather than void* as the type for the data, we'll need to use intptr_t or uintptr_t which should be the same size.   Java, for instance, stores pointers as the long data type.  But most wrappers can handle void* because it is such a common C idiom.

The key thing I am trying to determine is what happens on GCed systems.   For instance in the JVM world, the garbage collector can move objects.  So if there is no way to somehow track the changing pointer, then the C code will call the callback with an invalid pointer and possibly crash the VM.   While, to some extent, this is the responsibility of the application/wrapper, I want to make sure that there is a very clear explanation of how to do it right or have code that does it right.

Again, thanks for this contribution!

kyle-github:
Also, probably the name should be plc_tag_set_callback_ex() to make it clear that this is an extended version of the callback system.

I would also want to add @timyhac's request of having a PLCTAG_EVENT_CREATE_COMPLETE event too and that will require more tweaking.  That said, we can just add the event type but not generate it yet.

kyle-github:
I raised issue [#235](https://github.com/libplctag/libplctag.NET/issues/235) with the .Net team.

Joylei:
I'm good at .Net. It's not a problem for GCed systems. 
In .Net, there are various ways to pin memory that prevent it from moving.
- Allocate unmanaged buffer with `Marshal.AllocHGlobal`
- Use `fixed` keyword
- use `GCHandle`
- use other methods of `Marshal`

I believe there must be a similar mechanism for Java.

kyle-github:
Thanks for the response @Joylei.   I got the same response about C#/.Net from that wrapper team.   Java is going to be more of a challenge, however we can just include a reference to the original Tag object within the Callback instance so we do not need the passed reference.

What data type is preferable to .Net?   `void*` or `intptr_t`?

Joylei:
It's common to use `void*`. Anyway, it's up to the native library.

kyle-github:
Thanks, @Joylei!  I have moved this to the prerelease branch.   I do all PRs (mostly, in emergencies I'll commit to release) to that branch.   That is where testing is done.   It gives me a chance to catch other merge problems and test result problems.

I am going to try to get this merged this weekend.  I will do some testing and wiki additions first.

kyle-github:
This is going to end up as version 2.5.0 because it is an API change (SemVer).

kyle-github:
@Joylei  I have merged this into the prerelease branch.  I fixed up how callbacks are called to conform to the C standard.  You cannot "legally" cast a function pointer from a void pointer.   There are platforms where that won't work.   C, however, does let you cast a function pointer to another function pointer.  And due to the way that C marshals arguments, a caller can pass more arguments than a callee will use without danger.   

I added a few tests and fixed up a couple other things along the way (unrelated).

Please test and make sure it passes your internal tests!  If it looks good I will push this out as version 2.5.0.  The new minor version is due to the fact that this has an API change.  I still need to update the API wiki.

Issue #338, opened 2022-01-13
Status is still pending when read/write completed events are raised


Joylei:
I did multithreading bench testing and I found in the version v2.4.9 there are chances that status is still pending when read/write completed events are raised.
Then I did more testing with previous versions.
It seems that it happens from v2.4.5.

- platform: windows 10
- protocol=ab-eip&plc=controllogix

kyle-github:
Hi, can you give me more information about which status and the actions your program is taking in what order?

Do you mean the status as returned in a callback or the status from plc_tag_status()?

Joylei:
yes, I mean the status as returned in a callback.

I wrote the code in rust.
the async code is that:
- acquire a mutex lock for tag_id (in rust), ensure only one action can be performed for a tag. since it's multithreading, different tags have different mutex locks.
- read a tag: plc_tag_read(tag_id, 0)
- resume if a callback is received and the event is read completed,return the status from callback. because there is no way to pass in user data to a callback, there is a global state to track the callback, but it's lock protected.
- assert(status!=pending)
- do something and release the lock

the logic is reliable in previous versions. but it breaks now.

kyle-github:
Can you capture log output from this and attach it to this issue?   Set the logging level to 4.   This will dump out (on stderr) a lot of information about what the library is doing internally.   If this is a command line tool, you can redirect the log:

```text
C:\my_tool.exe > log.txt 2>&1
```

With the extra locking you are doing, I am not sure what is happening here.   I have a similar test in C and I will instrument it to check the status to see if I see anything like this.

Joylei:
please find the attached log.
[log.txt](https://github.com/libplctag/libplctag/files/7867643/log.txt)

```txt
2022-01-14 11:32:24.754 thread(2) tag(0) DETAIL check_read_status_connected:1418 type byte = 196 (c4)
thread 'main' panicked at '2022-01-14 11:32:24.754 thread(2) tag(0) INFO check_read_status_connected:1479 Got 4 bytes of data
assertion failed: !status.is_pending()', E:\Dev\public\plctag-rs\crates\async\src\op.rs2022-01-14 11:32:24.754 thread
```

It seems libplctag works fine.   let me check what happened with my code.

another log
[log.txt](https://github.com/libplctag/libplctag/files/7867713/log.txt)

Joylei:
[log.txt](https://github.com/libplctag/libplctag/files/7867790/log.txt)

the args of callback of libplctag returns status=1
```
[...]

PR #350, opened 2022-02-05
Fix event status

kyle-github:
Attempt to clean up event status.   Still looks like there might be event duplication.   Also looks like there will be some multi-threaded race conditions on event clean up.   Should be harmless (i.e. won't crash anything) but it is not clear if the results will be accurate.

### DIFF
--- a/src/examples/test_callback.c
+++ b/src/examples/test_callback.c
@@ -1,5 +1,5 @@
 /***************************************************************************
- *   Copyright (C) 2020 by Kyle Hayes                                      *
+ *   Copyright (C) 2022 by Kyle Hayes                                      *
  *   Author Kyle Hayes  kyle.hayes@gmail.com                               *
  *                                                                         *
  * This software is available under either the Mozilla Public License      *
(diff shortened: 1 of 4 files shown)


kyle-github:
@Joylei, @timyhac could you please give the prerelease branch a test.  This should be:

- generating a PLCTAG_EVENT_CREATED in most cases, though sometimes it will be synthesized based on another event firing.  I need to test more with special tags like @tags and @udt/.
- (more) correct status handling.  I am not seeing the same PENDING status on read completion that @Joylei  was finding anymore.

Caveats:

- it looks like I am generating some events twice.  Particularly read complete seems to fire more than once.  This may be an artifact of mixing output to stderr and stdout.
- I need to really dig into the code paths to make sure that there isn't any dangerous (i.e. crashing) race condition.

Joylei:
@kyle-github 

It's still not correctly handled.

Because the code will not reset `event_read_complete` and other similar flags if there is no callback.
We know that there is a read operation when initializing a tag. so the status is `PLCTAG_STATUS_OK` when the read is finished, and `tag_raise_event` is called, then we have `event_read_complete=1`.
If we register a callback and call `plc_tag_read` after that, the status is changed to `PLCTAG_STATUS_PENDING`. the `READ_STARTED` event is triggered, but since we have `event_read_complete=1`, it will immediately trigger `READ_COMPLETED` event. 

In a word, if we have some of these flags that was set to 1, then we register a callback, we'll get duplicated or wrong events we do not want.

The possible fix would be: do not set `tag->event_read_complete = 1` and other similar flags in `tag_raise_event` if there is no callback.

kyle-github:
Taking the conversation back over to #338.

PR #372, opened 2022-07-01
Add null terminator to tag->name string

### LINKED ISSUE #371: Possible non-null terminated string
Hi,

There's a possibility that the `str_copy` function at line [138](https://github.com/libplctag/libplctag/blob/release/src/protocols/system/system.c#L138) in `system_tag_create` function of `src/protocols/system/system.c` could create a non-null terminated `tag->name` string which can in turn lead to an out-of-bounds read (for example, if `strlen` is called on a non-null terminated string). This would be the case if the length of `name` string (which is copied to the `tag->name`) is longer than `MAX_SYSTEM_TAG_NAME`. The underlying reason for this is that `str_copy` uses `strncpy` to copy strings and as `strncpy` documentation mentions:

```
No null-character is implicitly appended at the end of destination if source is longer than num. Thus, in this case, destination shall not be considered a null terminated C string (reading it as such would overflow).
```

tinajohnson:
Fixes #371

### DIFF
--- a/src/protocols/system/system.c
+++ b/src/protocols/system/system.c
@@ -135,7 +135,8 @@ plc_tag_p system_tag_create(attr attribs, void (*tag_callback_func)(int32_t tag_
     tag->byte_order = &system_tag_byte_order;
 
     /* get the name and copy it */
-    str_copy(tag->name, MAX_SYSTEM_TAG_NAME, name);
+    str_copy(tag->name, MAX_SYSTEM_TAG_NAME - 1, name);
+    tag->name[MAX_SYSTEM_TAG_NAME - 1] = '\0';
 
     /* point data at the backing store. */
     tag->data = &tag->backing_data[0];


kyle-github:
Thanks for this.   This raises a larger issue: should the code simply fail in str_copy if the source string is larger than the buffer (modulo the null termination space)?  I'll have to think about that.   Unfortunately, C lets you ignore return values so even if I change the overall behavior of the function, the compiler is not going to help me find incorrect calls.

The forcing of null termination on line 139 where you put a null character into the last byte of the buffer is sufficient.  If the source string is shorter than the buffer size, then the underlying POSIX call will copy the string and null terminate it.  If it is longer, then the forced null terminator will truncate and terminate the string.

But this is where the above question comes up.   str_copy can change the string during the copy.   That is a non-obvious result and thus does not pass the principle of least surprize.

Let me think about this.   These wrappers were originally just minor code around underlying C library functions to make the APIs the same across platforms and to translate to library errors.   But in this kind of case, the semantics are not clean.

kyle-github:
Ensure that strings are null-terminated.

PR #389, opened 2022-10-15
Fix modbus test server

kyle-github:


kyle-github:
The pymodbus library had some API changes with version 3.0.  This fixes the test code so that it runs with that new API version.

Issue #451, opened 2024-03-26
String Write is not working for Omron NJ/NX Controller


salma4791:
Write operation for string datatype is not working. it is giving PLCTAG_ERR_TOO_LARGE while setting the string using plc_tag_set_string API. Observed that the string capacity is less than the string length due to which it is throwing this error.

ptsOSL:
Have you seen [https://github.com/libplctag/libplctag/issues/431]? There might be some information there to help you. I have managed to get writing to strings on the Omron NJ working, however in addition to correctly setting the additional tag attributes, I had to do two other things. First, I had to manually set the size of the libplctag tag buffer using `plc_tag_set_size()`, this is because I was trying to write a longer string to the PLC STRING variable than was currently stored in the PLC. This meant that the libplctag tag buffer was too small to store the data I wanted to write to the PLC, I think this might be your problem. Another issue I had was that libplctag forces padding to an even number of bytes, you may encounter this problem if you fix the first issue. There is more detail about this on the linked issue. 

Would it be appropriate for me to create another tag string attribute which allows the user to enable/disable having this string padding and create a pull request @kyle-github?

salma4791:
Yes i have seen the string issue hub but the plc_tag_set_size() is still not helping me because for Omron NJ/NX , the string datatype is appended with null byte. Additionally i have figured out that the length of the string while writing is not calculated properly.

ptsOSL:
I have created a pull request to fix the issue with the appended null byte, I understand that this is necessary for some other PLCs, but causes problems with Omron. What is the issue with the string length not being calculated properly? Is it the length of the read string or the calculated length of the string which is to be written to the PLC?

kyle-github:
I think @ptsOSL's PR is going to fix this.  If not, it will at least be a better base to start from.

kyle-github:
Please reopen once @ptsOSL's fix is out and it does not solve the problem.

Issue #431, opened 2023-07-20
Reading strings fields in structures on Omron NJ


exzombie:
Looking at the code, the `plc_tag_get_string()` function appears to assume that the first two bytes at the provided offset contain the string length. This does not appear to always be the case for Omron NJ. When the tag is `MyStruct.StringField` and `plc_tag_get_string()` is called with offset=0, the entire string is returned. However, when the tag is `MyStruct` and `plc_tag_get_string()` is called with the byte offset of the field, the first two characters are missing. Calling `plc_tag_get_raw_bytes()` with the same offset returns the entire string.

The Omron manual documents that when an array of strings is fetched, the array elements will not be prefixed by size. The manual does not discuss the strings that are part of structures, but it would seem that the same applies. It seems to me that the fix, then, would be to only read the size if the offset provided to `plc_tag_get_string()` is zero.

In #430, you mentioned that Omron support is a contribution from a while ago. Is it possible that strings as structure and array members were simply not tested at the time? Or could it be that this is something that changed with time?

timyhac:
There are some attributes in the tag string that should be able to configure this: https://github.com/libplctag/libplctag/wiki/Tag-String-Attributes#optional-generic-string-attributes

[It looks like in the case of Omron, `str_is_counted` defaults to `true` ](https://github.com/libplctag/libplctag/blob/4793d92893906e80fcfc67d4c4ed87fd37d81df8/src/protocols/ab/eip_cip.c#L196)but you suggest should default to `false` - can you link to the source documents?

exzombie:
Not really. String tags *are* counted. It's just strings in structures and arrays that are not. For arrays, this is described in section 8-7-4 of [EtherNet/IP Port User's Manual](https://assets.omron.eu/downloads/manual/en/v9/w506_nj_nx-series_cpu_unit_built-in_ethernet_ip_port_users_manual_en.pdf). The explanation sucks: there is insufficient detail, structs are not discussed, and the examples have errors (`??` invalid data, really? I'm pretty sure strings are null-terminated)

I don't have a string array to test with right now, so I can't say for sure what the byte order is. But what I *can* say for sure is
- The tag `MySingleString` refers to a counted string.
- The tag `MyStringArray[1]` refers to a counted string.
- The tag `MyStringArray` contains *non-counted* strings.
- The tag `MyStruct.StringMember` refers to a counted string.
[...]

Issue #430, opened 2023-07-19
Reading subarrays on Omron NJ


exzombie:
IIUC, libplctag relies on fragmented reads to fetch large tags, and that it not supported on Omron NJ. Skimming through the code, it seems to me that automatically splitting large tags into smaller chunks to be read with non-fragmented reads (service code 0x4C) does not fit the libplc architecture very well; I'd be happy to learn that I'm wrong about this, though. In any case, currently, this splitting into smaller "subtags" must be done by the client. I think that's perfectly fine, but  then, libplctag should provide the means for the client to do so, which is not currently the case.

Specifically, large arrays need to be fetched in several chunks. Currently, libplctag offers two options: fetch the entire tag, or fetch single elements. The first option does not work when the array is larger than the max response size of 1996B, the latter option results in reads that are too small; even with packing enabled, the performance is not good.

A better solution is to fetch the array in several chunks, larger than a single element but smaller than the max payload size. Omron NJ supports that with service code 0x4C, based on their documentation. In libplctag, this is done by specifying the number of elements for the tag. For example, to read elements 16-24, the tag name would contain `name=MyArray[16]&elem_count=9`.

However, for some reason, libplctag explicitly disables such use. In `ab_common.c`, we have
```
    switch(tag->plc_type) {
    case AB_PLC_OMRON_NJNX:
        if (tag->elem_count != 1) {
            tag->elem_count = 1;
            pdebug(DEBUG_WARN,"Attribute elem_count should be 1!");
        }
```
I wonder why this is so. I see no reason for this limitation. I've looked at #171 which introduced this, but the description is unclear to me. I tried removing the line which forces the element count to 1, everything appears to work. When the starting element and the count are specified properly, the read works. If they are specified incorrectly, the PLC returns `PLCTAG_ERR_BAD_PARAM`.

So, I propose to remove this limitation. Unless there is something I missed?

kyle-github:
Hi, thanks for using the library.   I don't have an Omron on which to test.   If you have one, I would be very happy to experiment to find out what to do!

The original Omron code was donated by a user. From what we could determine, the entire array (at least that part that fits) was returned regardless of what was placed in the element count.  At least on some of the tests that were done, an error was returned from the PLC when the element count was _not_ one (1).

As to the size of the returned result, while the total result size appears to be negotiated to be about 2k, each individual return was limited to something much smaller.

[...]

PR #455, opened 2024-04-12
Changes for improved Omron NJ/NX compatibility

ptsOSL:
This pull request relates to issues #430, #431 and possibly #451. The two primary things which this PR addresses are the hard limit on only being able to write a 1 single element to an array at once, and forced padding to an even number of bytes when writing strings. 

With regards to writing to multiple elements, I have been writing to multiple elements of an array on an Omron NJ with no problems. Only if you try and write to so many elements that you exceed the 1996 byte limit will there by a problem, but this is properly reported by the library and seems like fine behaviour to me.

With regards to the string padding, I thought that the best way to handle this would be to add a new string attribute which switches on/off padding to an even number of bytes. This defaults to on for all PLC types except Omron. One slight concern that I have with my implementation is that this also effects padding of other datatypes. However I think all datatypes except strings are an even number of bytes, even WORDS which are sent as two bytes? Anyway, if there is a way to check for the the datatype then let me know. I have also tested writing to a struct containing an odd number of bytes due to an odd sized string and this also works fine without the even byte padding.

I have tested this implementation on Omron NJ and it fixes the issue where for example a 5 byte string is sent as 6 bytes due to the even padding which makes the actual size contradict the string length specified. I have not tested on any other PLC types, but as it defaults to the old behaviour, there should be no effect.

Let me know what you think, or if you want any changes!

### DIFF
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -3980,7 +3980,7 @@ int set_tag_byte_order(plc_tag_p tag, attr attribs)
         use_default = 0;
     }
 
-    if(attr_get_str(attribs, "str_pad_to_16_bits", NULL) != NULL) {
+    if(attr_get_str(attribs, "str_pad_to_multiple_bytes", NULL) != NULL) {
         use_default = 0;
     }
 
@@ -4189,14 +4189,15 @@ int set_tag_byte_order(plc_tag_p tag, attr attribs)
             }
         }
 
-        /* Should we pad the string to an even number of bytes. Doing this causes issues when writing OmronNJ strings, 
-            required for certain AB PLCs*/
-        if(attr_get_str(attribs, "str_pad_to_16_bits", NULL)) {
-            str_param = attr_get_int(attribs, "str_pad_to_16_bits", 0);
-            if(str_param == 0 || str_param == 1) {
-                tag->byte_order->str_pad_to_16_bits = (unsigned int)str_param;
+        /* Should we pad the string to a multiple of 1 (no padding), 2, or 4 bytes. Adding padding causes issues when writing OmronNJ strings, 
+            2 byte padding is required for certain AB PLCs*/
+        if(attr_get_str(attribs, "str_pad_to_multiple_bytes", NULL)) {
+            str_param = attr_get_int(attribs, "str_pad_to_multiple_bytes", 0);
+            if(str_param == 0 || str_param == 1 || str_param == 2 || str_param == 4) {
+                if (str_param == 0) {str_param=1;} /* Padding to 0 bytes doesnt make much sense, so we overwride to 1 byte which means no padding */
+                tag->byte_order->str_pad_to_multiple_bytes = (unsigned int)str_param;
             } else {
-                pdebug(DEBUG_WARN, "Tag string attribute str_pad_to_16_bits must be missing, 0, or 1!");
+                pdebug(DEBUG_WARN, "Tag string attribute str_pad_to_multiple_bytes must be missing, 1, 2 or 4!");
                 return PLCTAG_ERR_BAD_PARAM;
             }
         }
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -3980,6 +3980,10 @@ int set_tag_byte_order(plc_tag_p tag, attr attribs)
         use_default = 0;
     }
 
+    if(attr_get_str(attribs, "str_pad_to_16_bits", NULL) != NULL) {
+        use_default = 0;
+    }
+
     /* if we need to override something, build a new byte order structure. */
     if(!use_default) {
         const char *byte_order_str = NULL;
@@ -4181,6 +4185,18 @@ int set_tag_byte_order(plc_tag_p tag, attr attribs)
                 tag->byte_order->str_pad_bytes = (unsigned int)str_param;
             } else {
                 pdebug(DEBUG_WARN, "Tag string attribute str_pad_bytes must be missing, 0, or positive!");
+                return PLCTAG_ERR_BAD_PARAM;
+            }
+        }
+
+        /* Should we pad the string to an even number of bytes. Doing this causes issues when writing OmronNJ strings, 
+            required for certain AB PLCs*/
+        if(attr_get_str(attribs, "str_pad_to_16_bits", NULL)) {
+            str_param = attr_get_int(attribs, "str_pad_to_16_bits", 0);
+            if(str_param == 0 || str_param == 1) {
+                tag->byte_order->str_pad_to_16_bits = (unsigned int)str_param;
+            } else {
+                pdebug(DEBUG_WARN, "Tag string attribute str_pad_to_16_bits must be missing, 0, or 1!");
                 return PLCTAG_ERR_BAD_PARAM;
             }
         }
--- a/src/protocols/ab/eip_cip.c
+++ b/src/protocols/ab/eip_cip.c
@@ -174,6 +174,7 @@ tag_byte_order_t logix_tag_byte_order = {
     .str_is_fixed_length = 1,
     .str_is_zero_terminated = 0,
     .str_is_byte_swapped = 0,
+    .str_pad_to_16_bits = 1,
 
     .str_count_word_bytes = 4,
     .str_max_capacity = 82,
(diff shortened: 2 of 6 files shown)


kyle-github:
I am trying to run the approval workflow to see if this passes the base tests.   Sorry for the delay in response here.

Are you OK with licensing this with a dual MPL/LGPL license?   I want to make sure that licensing for the whole library stays as simple as possible.  

I may have more comments as I go through the code changes.

kyle-github:
BTW, data types like BOOL and SINT/USINT are single bytes.

kyle-github:
Looks like all the base checks passed.  I will move this over to the prerelease branch because I merge all changes there first, update the version and then merge over to the release branch for a final run and packaging.

kyle-github:
I did a check of the code.  I like most of it, thanks!

On the padding, I think I might want to do a slightly different direction.   Rather than specifically having a 16-bit assumption on the string length, it might be better to have a different attribute, str_pad_to_multiple_bytes, or something (that name is not great) where you can put values like 1 for byte-level padding (i.e. none), 2 for 16-bit word padding, 4 for 32-bit word padding (I think this existing in some protocols) etc.

Let me explore that a bit. It might take a couple of days.

kyle-github:
Oops, I was looking at the wrong spot in GitHub, you've already PRed this for the prerelease branch.  Sorry!

ptsOSL:
> BTW, data types like BOOL and SINT/USINT are single bytes.

Thanks, I've realised that I was confusing a padding byte with being a part of the datatype

src/protocols/ab/eip_cip.c
```diff
@@ -996,6 +999,7 @@ int build_write_request_connected(ab_tag_p tag, int byte_offset)
     ab_request_p req = NULL;
     int multiple_requests = 0;
     int write_size = 0;
+    int pad_to_even_bytes = 1;
```
kyle-github: Is there a reason you were inconsistent with the naming?   I actually like this better than the 16 name...  Just wondering if I am missing something here.

ptsOSL: Nope, that was just a mistake on my behalf. I was planning to go ahead with your idea of changing the parameter to str_pad_to_multiple_bytes anyway, as that seems like an improvement and will be quite simple to implement.

kyle-github:
I move the base branch to omron_updates.  This branch is forked off of prerelease.

src/lib/lib.c
```diff
@@ -4189,14 +4189,15 @@ int set_tag_byte_order(plc_tag_p tag, attr attribs)
...
-        if(attr_get_str(attribs, "str_pad_to_16_bits", NULL)) {
-            str_param = attr_get_int(attribs, "str_pad_to_16_bits", 0);
-            if(str_param == 0 || str_param == 1) {
-                tag->byte_order->str_pad_to_16_bits = (unsigned int)str_param;
+        /* Should we pad the string to a multiple of 1 (no padding), 2, or 4 bytes. Adding padding causes issues when writing OmronNJ strings, 
+            2 byte padding is required for certain AB PLCs*/
+        if(attr_get_str(attribs, "str_pad_to_multiple_bytes", NULL)) {
+            str_param = attr_get_int(attribs, "str_pad_to_multiple_bytes", 0);
+            if(str_param == 0 || str_param == 1 || str_param == 2 || str_param == 4) {
+                if (str_param == 0) {str_param=1;} /* Padding to 0 bytes doesnt make much sense, so we overwride to 1 byte which means no padding */
```
kyle-github, line 4197: Maybe just make 0 an error?

src/protocols/ab/eip_cip.c
```diff
@@ -1090,14 +1090,17 @@ int build_write_request_connected(ab_tag_p tag, int byte_offset)
...
-    pad_to_even_bytes = tag->byte_order->str_pad_to_16_bits;
-    if (pad_to_even_bytes == 1){
-        if (write_size & 0x01) {
-            *data = 0;
-            data++;
+    str_pad_to_multiple_bytes = (int)tag->byte_order->str_pad_to_multiple_bytes;
+    if ((str_pad_to_multiple_bytes == 2 || str_pad_to_multiple_bytes == 4) && write_size != 0){
+        if (write_size % str_pad_to_multiple_bytes !=0){
+            int pad_size = str_pad_to_multiple_bytes-(write_size % str_pad_to_multiple_bytes);
+            for (int i =0; i<pad_size; i++){
```
kyle-github, line 1100: Should check that we have buffer space.

src/protocols/ab/eip_cip.c
```diff
@@ -1245,17 +1248,21 @@ int build_write_request_unconnected(ab_tag_p tag, int byte_offset)
...
```
kyle-github, line 1258: Again, need to check for running over the buffer.

kyle-github:
The latest looks really good.  Thanks for that!   The only potential problem I see is when the padding is added, there is no check to make sure we are not going over the internal buffer limit.   This raises the issue that using plc_tag_set_size() is not going to work well with strings that are padded because you would need to know the padding.  And that is not exposed.   Let me think about this.   Might be a bigger issue.

src/lib/lib.c
ptsOSL: Sure

src/protocols/ab/eip_cip.c
ptsOSL: This does seem like something that should be happening, but Im not sure how to check the size here and dont see it being checked anywhere else in this function? Is it req->request_capacity im looking for?

kyle-github: Hmm, I am going to have to look at the surrounding code.   The variable "write_size" does hold the amount we are going to write, but that could be constrained by things like a full request buffer.

Also I just realized that I am not sure that this padding is applicable only to strings.   The original code did not care and just padded everything.  That's a problem.   Phwew if this needs to be type-aware... yuck.

kyle-github:
I found this about Omron:

- Omron’s NJ structure types use the natural alignment of their members, up to 64-bit, then pad the structure to the largest alignment of it the members. Which then becomes the structure’s
natural alignment when nesting in other structures. Named booleans in NJ structures occupy
and align to a complete 16-bit word of their own.
- Omron’s CJ structure types use only 16-bit alignment, even for bytes and strings. Consecutive
named booleans are packed into 16-bit words together, and in this library, must be hosted in a
16-bit word.

So alignment is different between CJ and NJ PLCs.  And BOOLs in UDTs apparently map to single bits in a 16-bit word, not an 8-bit word as in Control/CompactLogix.   Ooph.   Omron doesn't want to make this easy.

ptsOSL:
Well Ill certainly agree that Omron doesnt like to make things easy. With the boolean alignment, my understanding is that each boolean aligns to 16-bits when requested individually or within a UDT, even if there are consecutive booleans in the UDT. Within an array however, consecutive booleans will pack together, ie you can fit 16 booleans in 16 bits. Memory alignment is explained in a good level of detail in https://edata.omron.com.au/eData/NJ/W506-E1-24.pdf starting on page 550.

The 64-bit alignment around structures based on their largest member is correct.

I have written some code for an OmronNJ which can map a user supplied list of dtypes within a UDT to the offsets to the data within the UDT (including size of dtype and padding/alignment of the dtypes, structures and arrays), and it was quite painful.

src/protocols/ab/eip_cip.c
ptsOSL: I have been thinking about this a bit, originally I did only intend this to affect writing strings but it will affect all writes. I have been testing with these changes and not noticed any problems with writing other datatypes on OmronNJ. It seems that at least for Omron, you do not need to write the correct number of padding 0's for the alignment of the datatype. If this was necessary then you would need to add different numbers of 0's for different datatypes (which have different padding rules) and previously the code always padded to an even number of bytes. The only time there is an issue is when we do add padding (for writing strings).

Do different dtypes have different alignments in other PLCs or is this just an Omron thing? Is the idea of this padding to match the alignment of the data or is it just intended to give an even number of bytes? If we removed the padding altogether would writes fail for other PLC types?

If the user specifies this new str_pad_to_multiple_bytes for a non string tag when the documentation will clearly state that it is only for strings, is that a problem we should be worried about? It might not be a problem that this code affects non string dtypes, as the tag which has str_pad_to_multiple_bytes enabled should only be reading a string.

kyle-github:
Let's just give this a try.   We can go around in circles, but if all the checks pass against my various PLCs and against your Omron, that should be good enough.

ptsOSL:
Thanks Kyle! Ill still be working on Omron for the next couple of months, so if I notice any issues ill let you know.

Issue #454, opened 2024-04-12
Message packing on OmronNJ/NX beyond max response size


ptsOSL:
I have been encountering a problem with packing of CIP requests when reading data from OmronNJ. libplctag is packing many read requests into a single request message, as I have requested. However, if the size of the response packet (the size of all of the data which is being read) is beyond the max packet size of 1996 bytes, then I receive error 0x1B, routing failure (response packet too large). This is mostly a problem when packing multiple large structures together. I am guessing this is only a problem on Omron as it does not support fragmented reads. Looking through the source code, I didn't see anywhere that libplctag checks to see if the combined size of the response message would fit in a single response packet, but i might have missed it? Would it make sense to individually read each item that is to be packed when the tags are initialized to see if they can be packed together? This is something the user can do, however it seems more like something libplctag should do, but I might be misunderstanding something. Cheers

kyle-github:
I don't have an Omron, so take what I say with a grain of salt.   From what I understand, Omron decided to behave differently than Rockwell in some key areas:

- packet size negotiation works, but only up to about 1900 bytes.
- When retrieving a tag, the count seems to be ignored and (possibly?) may be refused if it is not 1.   
- When retrieving an array tag, the entire tag is returned instead of just the requested elements (this was reported but I cannot verify).
- There is a limit on the size of the returned data.   I think it is around 256 or 512 bytes.   Anything larger than that seems to fail.   
- There are additional CIP commands that Omron uses to get arrays (or array elements?) that are larger than the above limit.
- Omron does _NOT_ support the 0x06 status code which means "partial result" for Rockwell.  Rockwell uses this to return partial results and then uses the offset field in the request to get the next chunk of data.

One of the results of all this is that you can request several tags at once, but if the results do not fit into the 1900 byte result packet, the whole bundled request seems to fail.  I think this is what you are hitting.

So while Omron is using CIP it is not using the Rockwell commands.   Unfortunately, I have not found a good source of documentation for Omron on this.   There is a [Python library](https://github.com/aphyt/aphytcomm) that supports Omron with CIP that could be a source of ideas.

ptsOSL:
So I have been developing with libplctag for OmronNJ for a few months now and some of the behaviour you describe differs to what I have seen on my Omron NJ:

- I believe the maximum packet size is 1996, its certainly around that number anyway, it is not actually specified in the manual
[...]

PR #461, opened 2024-05-21
Added response packet size calculation and checking when adding reque…

ptsOSL:
…sts to a msp request.

Hi Kyle, these changes are to address [issue 454]( https://github.com/libplctag/libplctag/issues/454). What I have done is modified the code which adds multiple packets into a multiple service packet (msp) read/write request, to keep track of the size of the data which is expected to be returned. After each tag is read once, we now know the size of this data (assuming it hasnt changed between reads). I have used this size as well as adding some extra space to account for the embedded cip headers, to track how big the response packet is expected to be. For PLCs where fragmented reads is not supported (ie omron), all of the response data must fit into a single packet, these changes should stop the response data being too big for a single packet.

Currently there is no protection against this, other than defaulting allow_packing to be switched off, so these changes will add some protection. However if data grows between reads for some reason (this can be the case for Omron strings), then this protection will fail and there will be an error returned. So it would be a good idea to add this caveat to the documentation somewhere.

PLCs which support fragmented reads should be unaffected by these changes. Could you check that I have correctly identified which PLCs support fragmented reads? I am not familiar with the alan bradley PLCs. Also I have not tested whether this protection works for other PLCs which support message packing but not fragmented reads (i think it should). I have tested with my Omron NJ and have been able to pack 11 read requests into a msp and successfully read a single 1994 byte response msp.

Let me know what you think, 
Cheers Phil

### DIFF
--- a/src/protocols/ab/eip_cip.c
+++ b/src/protocols/ab/eip_cip.c
@@ -488,6 +488,13 @@ int build_read_request_connected(ab_tag_p tag, int byte_offset)
 
     req->allow_packing = tag->allow_packing;
 
+    /* set the response size to the size of data from the previous read */
+    req->response_capacity = tag->size;
+
+    /* if this is the first read of the tag then we do not know the size of the response data and cannot use packing unless the plc supports fragmented reads*/
+    req->first_read = tag->first_read;
+    req->supports_fragmented_read = tag->supports_fragmented_read;
+
     /* add the request to the session's list. */
     rc = session_add_request(tag->session, req);
 
@@ -626,6 +633,13 @@ int build_read_request_unconnected(ab_tag_p tag, int byte_offset)
 
     /* allow packing if the tag allows it. */
     req->allow_packing = tag->allow_packing;
+
+    /* set the response size to the size of data from the previous read */
+    req->response_capacity = tag->size;
+
+    /* if this is the first read of the tag then we do not know the size of the response data and cannot use packing unless the plc supports fragmented reads*/
+    req->first_read = tag->first_read;
+    req->supports_fragmented_read = tag->supports_fragmented_read;
 
     /* add the request to the session's list. */
     rc = session_add_request(tag->session, req);
@@ -776,6 +790,9 @@ int build_write_bit_request_connected(ab_tag_p tag)
 
     /* allow packing if the tag allows it. */
     req->allow_packing = tag->allow_packing;
+
+    /* sets the expected size of the response message */
+    req->response_capacity = 0;
 
     /* add the request to the session's list. */
     rc = session_add_request(tag->session, req);
@@ -965,6 +982,9 @@ int build_write_bit_request_unconnected(ab_tag_p tag)
     /* allow packing if the tag allows it. */
     req->allow_packing = tag->allow_packing;
 
+    /* sets the expected size of the response message */
+    req->response_capacity = 0;
+
     /* add the request to the session's list. */
     rc = session_add_request(tag->session, req);
 
@@ -1112,6 +1132,9 @@ int build_write_request_connected(ab_tag_p tag, int byte_offset)
 
     /* allow packing if the tag allows it. */
     req->allow_packing = tag->allow_packing;
+
+    /* sets the expected size of the response message */
+    req->response_capacity = 0;
 
     /* add the request to the session's list. */
     rc = session_add_request(tag->session, req);
@@ -1293,6 +1316,9 @@ int build_write_request_unconnected(ab_tag_p tag, int byte_offset)
 
     /* allow packing if the tag allows it. */
     req->allow_packing = tag->allow_packing;
+
+    /* sets the expected size of the response message */
+    req->response_capacity = 0;
 
     /* add the request to the session's list. */
     rc = session_add_request(tag->session, req);
--- a/src/protocols/ab/session.c
+++ b/src/protocols/ab/session.c
@@ -1380,7 +1380,9 @@ int process_requests(ab_session_p session)
     ab_request_p request = NULL;
     ab_request_p bundled_requests[MAX_REQUESTS] = {NULL};
     int num_bundled_requests = 0;
-    int remaining_space = 0;
+    int remaining_request_space = 0;
+    int remaining_response_space = 0;
+    int allow_packing = 0;
 
     debug_set_tag_id(0);
 
@@ -1408,28 +1410,42 @@ int process_requests(ab_session_p session)
             /* if there are still requests after purging all the aborted requests, process them. */
 
             /* how much space do we have to work with. */
-            remaining_space = session->max_payload_size - (int)sizeof(cip_multi_req_header);
+            remaining_request_space = session->max_payload_size - (int)sizeof(cip_multi_req_header);
+
+             /* -2 bytes for the msp (multi service packet) number of packets and -4 bytes for the cip response header,
+              * we later subtract 2 bytes for each packet, this is to allow for the INT offset to the packet within the msp */
+            remaining_response_space = session->max_payload_size - 2 - 4;
 
             if(vector_length(session->requests)) {
                 do {
                     request = vector_get(session->requests, 0);
 
-                    remaining_space = remaining_space - get_payload_size(request);
+                    remaining_request_space = remaining_request_space - get_payload_size(request);
+
+                    /* calculate if all of the response data from the packed requests will fit into a single response packet. 
+                     * the -8 bytes is the maximum padding between packets, the -2 bytes is from the offset integer which stores the
+                     * offset to this packet */
+                    remaining_response_space = remaining_response_space - request->response_capacity - 8 - 2; 
+
+                    /* The tag must have packing enabled and the plc must either support fragmented reads or this tag must have been read 
+                     * before, so that its response size is known */
+                    allow_packing = request->allow_packing && (request->supports_fragmented_read || !request->first_read);
 
                     /*
                      * If we have a non-packable request, only queue it if it is the first one.
-                     * If the request is packable, keep queuing as long as there is space.
+                     * If the request is packable, keep queuing as long as there is space in both the 
+                     * request packet and the response packet if fragmented reads is not supported.
                      */
 
-                    if(num_bundled_requests == 0 || (request->allow_packing && remaining_space > 0)) {
-                        //pdebug(DEBUG_DETAIL, "packed %d requests with remaining space %d", num_bundled_requests+1, remaining_space);
+                    if(num_bundled_requests == 0 || (allow_packing && remaining_request_space > 0 && (remaining_response_space >= 0 || request->supports_fragmented_read))) {
+                        //pdebug(DEBUG_DETAIL, "packed %d requests with remaining request space %d and remaining response space %d", num_bundled_requests+1, remaining_request_space, remaining_response_space);
                         bundled_requests[num_bundled_requests] = request;
                         num_bundled_requests++;
 
                         /* remove it from the queue. */
                         vector_remove(session->requests, 0);
                     }
-                } while(vector_length(session->requests) && remaining_space > 0 && num_bundled_requests < MAX_REQUESTS && request->allow_packing);
+                } while(vector_length(session->requests) && remaining_request_space > 0 && num_bundled_requests < MAX_REQUESTS && allow_packing && (remaining_response_space >= 0 || request->supports_fragmented_read));
             } else {
                 pdebug(DEBUG_DETAIL, "All requests in queue were aborted, nothing to do.");
             }
--- a/src/protocols/ab/tag.h
+++ b/src/protocols/ab/tag.h
@@ -113,6 +113,7 @@ struct ab_tag_t {
     int offset;
 
     int allow_packing;
+    int supports_fragmented_read;
 
     /* flags for operations */
     int read_in_progress;
--- a/src/protocols/ab/ab_common.c
+++ b/src/protocols/ab/ab_common.c
@@ -218,27 +218,32 @@ plc_tag_p ab_tag_create(attr attribs, void (*tag_callback_func)(int32_t tag_id, 
     case AB_PLC_PLC5:
         tag->use_connected_msg = 0;
         tag->allow_packing = 0;
+        tag->supports_fragmented_read = 0;
         break;
 
     case AB_PLC_SLC:
         tag->use_connected_msg = 0;
         tag->allow_packing = 0;
+        tag->supports_fragmented_read = 0;
         break;
 
     case AB_PLC_MLGX:
         tag->use_connected_msg = 0;
         tag->allow_packing = 0;
+        tag->supports_fragmented_read = 0;
         break;
 
     case AB_PLC_LGX_PCCC:
         tag->use_connected_msg = 0;
         tag->allow_packing = 0;
+        tag->supports_fragmented_read = 0;
         break;
 
     case AB_PLC_LGX:
         /* default to requiring a connection and allowing packing. */
         tag->use_connected_msg = attr_get_int(attribs,"use_connected_msg", 1);
         tag->allow_packing = attr_get_int(attribs, "allow_packing", 1);
+        tag->supports_fragmented_read = 1;
         break;
 
     case AB_PLC_MICRO800:
@@ -248,6 +253,7 @@ plc_tag_p ab_tag_create(attr attribs, void (*tag_callback_func)(int32_t tag_id, 
 
         /* Micro800 cannot pack requests. */
         tag->allow_packing = 0;
+        tag->supports_fragmented_read = 1;
         break;
 
     case AB_PLC_OMRON_NJNX:
@@ -258,6 +264,7 @@ plc_tag_p ab_tag_create(attr attribs, void (*tag_callback_func)(int32_t tag_id, 
          * whether the results will fit or not.
          */
         tag->allow_packing = attr_get_int(attribs, "allow_packing", 0);
+        tag->supports_fragmented_read = 0;
         break;
 
     default:
@@ -1187,9 +1194,3 @@ int check_write_request_status(ab_tag_p tag, ab_request_p request)
 
     return rc;
 }
-
-
-
-
-
-
--- a/src/protocols/ab/session.h
+++ b/src/protocols/ab/session.h
@@ -131,6 +131,10 @@ struct ab_request_t {
     /* used by the background thread for incrementally getting data */
     int request_size; /* total bytes, not just data */
     int request_capacity;
+    int response_capacity; /* Size of data we expect to be returned by this request */
+
+    int first_read;
+    int supports_fragmented_read;
     uint8_t *data;
 };
 


src/protocols/ab/eip_cip.c
```diff
@@ -488,6 +488,13 @@ int build_read_request_connected(ab_tag_p tag, int byte_offset)
 
     req->allow_packing = tag->allow_packing;
 
+    /* set the response size to the size of data from the previous read */
+    req->response_capacity = tag->size;
```
kyle-github, line 492: Actually... this won't work.  Some tags have more data than fits in a request.   In that case, the tag size is greater than the request buffer size.

src/protocols/ab/eip_cip.c
```diff
@@ -627,6 +634,13 @@ int build_read_request_unconnected(ab_tag_p tag, int byte_offset)
     /* allow packing if the tag allows it. */
     req->allow_packing = tag->allow_packing;
 
+    /* set the response size to the size of data from the previous read */
+    req->response_capacity = tag->size;
```
kyle-github, line 638: Same comment as previous.   The request response size is limited to what you can force in a single response packet (~4k for ControlLogix with newer network modules or a 1756-L8x CPU, ~500 bytes for other Rockwell CIP PLCs, and closer to 256 bytes for PCCC/CSP PLCs like PLC/5).  So I could have a tag with a size of 16k but only get back 4k at a time.

src/protocols/ab/session.c
```diff
@@ -1408,28 +1410,42 @@ int process_requests(ab_session_p session)
...
+                    remaining_request_space = remaining_request_space - get_payload_size(request);
+
+                    /* calculate if all of the response data from the packed requests will fit into a single response packet. 
+                     * the -8 bytes is the maximum padding between packets, the -2 bytes is from the offset integer which stores the
+                     * offset to this packet */
+                    remaining_response_space = remaining_response_space - request->response_capacity - 8 - 2; 
+
+                    /* The tag must have packing enabled and the plc must either support fragmented reads or this tag must have been read 
+                     * before, so that its response size is known */
+                    allow_packing = request->allow_packing && (request->supports_fragmented_read || !request->first_read);
```
kyle-github, line 1432: I am not sure why you are doing this logic.  Is this to make sure we know the size of the tag first?   Hmm, that makes sense.   Let me know if this is the right guess.

src/protocols/ab/tag.h
```diff
@@ -113,6 +113,7 @@ struct ab_tag_t {
     int offset;
 
     int allow_packing;
+    int supports_fragmented_read;
```
kyle-github, line 116: I probably should make these bit fields.  An int is a lot of wasted space.

kyle-github:
This looks pretty good but I am wondering if it makes sense just to copy the entire source tree for CIP into Rockwell and Omron trees.  That way it would be possible to do a lot more of these tweaks without all the extra logic.   We know, when the tag is created, what kind of PLC this is.

I also found out, on a PLC site, a while ago that Omron can have tags that are larger than the packet size.  In that case you need to read chunks of the tag at a time.  This is fairly straightforward with arrays but not as clear to me with large UDTs.   Do you know if this size restriction (or lack thereof) is true?  

For arrays, you would need to calculate the response size and then figure out how many elements that is and then step through the array with multiple requests.  I worked this out for PLC/5 a while ago in another branch and it was finicky but not that hard.

src/protocols/ab/eip_cip.c
ptsOSL: Yes, this wont work to fetch a single array or UDT which is larger than the response buffer size, it wasn't intended to do that. This PR was intended only to allow packing of many requests for which the response is less than the buffer size.

If the user requests a UDT/array which is too big for the buffer, they already get an easily understood "data too large" type error which I see as being fine. If  however they  enable packing and request many items which are smaller than the buffer size, they will still get a "too large" type error. This seems more like a bug to me and is what this PR is hoping to stop from happening.

There is a potential problem here if the tag->size says that the data will fit and then when it is actually read, the fetched data is larger than this for some reason.

src/protocols/ab/eip_cip.c
ptsOSL: I understood that fragmented reads like you say are available on some of the Rockwell PLCs, although I have no experience with them myself. This solution should be able to coexist with the fragmented reads that these PLCs do. It will not be as efficient as using fragmented reads, but for PLCs which dont support this, it will be an improvement over no packing, assuming we have a good way of working out how much data the PLC will return.

src/protocols/ab/session.c
ptsOSL: That's correct, this is to make sure that we either know the size of the response data we are requesting (so we know it will fit in the buffer), or that the PLC supports fragmented reads in which case it doesn't matter.

src/protocols/ab/tag.h
ptsOSL: Haha, yes that might be nice, I tried to make them bools but then realized that the Booleans dtype wasn't defined and didn't want to do that myself.

ptsOSL:
I think its probably not worth splitting the CIP source code into Rockwell and Omron but ill leave that decision up to you, I have now made most of the changes for Omron that I would like to see, I am hoping to implement a couple of the Omron specific TIME dtypes at some point, but ill leave that for another PR.

With Omron at least, it is straightforward to read chunks of UDTs by simply addressing individual items within the UDT. Such as structName.itemName. Ofcourse this would be a pain if you had a very large UDT with lots of basic dtypes arranged directly within the UDT rather than organised within sub-UDTs or sub-arrays. Also I can confirm that you cannot request an array beyond the 1996 byte restriction, I have tried and you will get a data size related error message. I have not tried with a UDT, or actually seen one bigger than 1996 bytes, but I am pretty certain you would get the same error.

I think being able to make a request of the driver to read an array beyond the size of the response buffer by automagically splitting it up would be useful, bit I don't think it is necessary and I don't intend to implement that here.

kyle-github:
I've been going over this again.   Maybe we are making this too hard?  Right now, all regular tags do a hidden first read that takes place while the tag creation process it happening.   There are three reasons I do this:

1. This gets the encoded data type of the tag.   This is needed if you want to write.
2. It gets the full size of the tag and thus pads out the buffer early.
3. It gets the current tag state so that if someone wants to just change a couple of parts of a big tag and then write it immediately they canl.

In order to do 1 and 2 we don't need to read the whole tag.  We only need to read one element.  So we can force the element count to 1 again, but just on first read.

For 3, I don't know that anyone is actually using it. This might be too much of a break.   

If we can just read one element of an array tag, we'll have a lot less data coming back in the most common case.  If the elements are large UDTs, we'd have the same problem anyway.

I am really tempted to do the above logic.  Thoughts?

BTW, the library currently negotiates the same packet size for both request and response.  So a request buffer is the same size no matter which "direction" we are going.   There is only one buffer in the request.

kyle-github:
Ah, hmm...  Reading the first one will not work with an array of strings unless they are come back all the same size.   Still might be worth it.

ptsOSL:
> If we can just read one element of an array tag, we'll have a lot less data coming back in the most common case. If the elements are large UDTs, we'd have the same problem anyway.

If I understand correctly you want to calculate the size of the array using the number of elements and the size of the first element? That sounds like a good idea to me, it will certainly speed up the initial reads and should be safe at least for basic datatypes, although its not necessary for this PR. Also I thought that an array of strings had to contain only strings of the same size, this is the case for Omron as far as I can tell (each string takes up the same amount of data in the array, but when requesting the strings directly you may be sent different amounts of data depending on what is stored in the string).

Maybe we are overthinking things, so to clarify, this PR was originally intended to be fairly simple. It waits for this initial read of each tag as you describe so that it can get the size of each tag. Then, if the user has enabled packing, before blindly packing as many read requests into a single multi service request, it will add up the size of the tag buffer for each tag being added to give a good guess at how much data it expects to receive in total. For PLCs which dont support fragmented reads, it checks that the total data expected is less than the max message size before adding the tag. If the data will not fit, it instead sends the current multi service request and starts a new one.

kyle-github:
The problem with reading the entire set of elements during a first read comes when you cannot fit many results into a packed response.   Setting up a lot of tags is already a very slow operation.   I have tests that set up 1000 tags and it takes a while even with partial results, packing etc.  But if I only get the first element, then I can pack even more of those things into one request/response.

So I think the read-first-element idea has some legs.   I am going to run with it in a few days and see how far I get.

kyle-github:
Thinking a little more about this:

- the total space for a request/response within a multi-request (service 0x0A to the Message Router) packet is:
    - the size of the full CIP request which is usually a CIP request header including the service code, the encoded tag name, a count of elements, and an optional offset in bytes (if fragmented read/write is supported).
    - a single INT (16-bit) offset in the 0x0A packet.
    - a static part shared for all tags.

Here's an example:

```
0a                         -- Multirequest service
02 20 02 24 01    -- sent to Class ID 2, Instance ID 1 (Message Router) 
1e 00                    -- 30 requests are packed in this packet
3e 00                   -- request zero starts at offset 62
56 00                   -- request one starts at offset 86
6e 00                   ...
86 00 
9e 00
b6 00 
ce 00 
e6 00 
fe 00 
16 01 
2e 01 
46 01 
5e 01 
76 01 
8e 01 
a6 01 
be 01 
d6 01 
ee 01 
06 02 
1e 02 
36 02 
4e 02 
66 02 
7e 02 
96 02 
ae 02 
c6 02 
de 02 
f6 02 
52 08 91 0c 54 65 73 74 42 69 67 41 72 72 61 79 28 00 21 00 00 00 00 00 -- first read request
52 08 91 0c 54 65 73 74 42 69 67 41 72 72 61 79 28 01 21 00 00 00 00 00  -- second read request
...
```

From this you can see that the fixed part of the 0x0A service request are the first 8 bytes.  After that, there is a list of the offsets, one per packed request.  So the total space required for a packet is:

8 + N*(2 + size(R[n])). where n goes from 1 to N.

On the response, the header is simpler.

```
8a 00 1e 00  -- response was good.  Code 0x1e means partial response.
1d 00  -- 29 responses in this packet 
3c 00 -- offset of the first response
c6 00 -- 2nd
50 01 ...
da 01 
64 02 
ee 02 
78 03 
02 04 
8c 04 
16 05 
a0 05 
2a 06 
b4 06 
3e 07 
c8 07 
52 08 
dc 08 
66 09 
f0 09 
7a 0a 
04 0b 
8e 0b 
18 0c 
a2 0c 
2c 0d 
b6 0d 
40 0e 
ca 0e 
54 0f 
d2 00 00 00 c4 00 11 00 00 00 12 00 00 00 13 00 00 00 34 00 00 00 15 00 00 00 16 00 00 00 17 00 00 00 18 00 00 00 cd 01 00 00 a9 00 00 00 a8 00 00 00 ad 00 00 00 ab 00 00 00 ab 00 00 00 ab 00 00 00 a6 00 00 00 aa 00 00 00 ac 00 00 00 ab 00 00 00 2c 01 00 00 2d 01 00 00 2e 01 00 00 30 01 00 00 2f 01 00 00 2b 01 00 00 2f 01 00 00 2f 01 00 00 2e 01 00 00 2f 01 00 00 75 01 00 00 78 01 00 00 7b 01 00 00 79 01 00 00 d2 00 00 00 c4 00 12 00 00 00 13 00 00 00 34 00 00 00 15 00 00 00 16 00 00 00 17 00 00 00 18 00 00 00 cd 01 00 00 a9 00 ...  -- first response
```

ptsOSL:
So to calculate the size of the response data. It is the 4 byte response header + 2 bytes containing the number of responses + 2 bytes for the offset to each response + the data for each response. So 6 + N*(2 + size(R[n])). where n goes from 1 to N

Which should be what my changes implement. The main assumption my changes make is that the size of the data is the same size as it was when the tag was last read. I dont think there is a good way around this, except to disallow packing of datatypes that can change size or adding warnings etc.

kyle-github:
That calculation looks right to me.  Are there data types that change size?   Strings maybe, but most of those have a fixed capacity (i.e. 82 chars).

ptsOSL:
It was just Omron strings that I was thinking of. Only the chars stored in the string up to and including \n are sent, not the entire string, this means that they can change size as the string inside changes.

ptsOSL:
Sooo, don't know how I didn't see this earlier, but the byte limit for Omron NJ is actually 1994 bytes not 1996 as coded insession.c. This explains why I could read a 1994 byte packed message, but 1996 was failing. Here is the extract from page 265 of the manual:

image

The NX701 seems to have a 8192 byte limit, but I don't have one to test it.

kyle-github:
Ooh, good catch.  I note that the negotiation using Forward Open is fine with 1996 bytes.   Do you want to update this PR? 

Also, since I released v2.6.0, you may need to rebase this.

kyle-github:
Can you do a test?   Set the number to something high.  Like 16k.  Run tag_rw2 against the PLC.  Look for the string "with packet size" in the logs.   That is, if it succeeds.  In some cases, newer PLCs will respond to a size that they do not support with one they will support.

I try to keep the numbers exact because there are still PLCs that just respond with "Nope" but do not tell the library what size packet they can take.

kyle-github:
I'm going to take a look at where this is all set.   The smaller packet size needs to be fixed too.

ptsOSL:
[log.txt](https://github.com/user-attachments/files/15912246/log.txt)

The PLC is reporting a maximum size of 1996 bytes which is in disagreement with the 1994 bytes stated in the manual. I have no idea why.

kyle-github:
Yeah, I see that with other Omron PLCs that people are testing with.  

None of these manuals seem accurate.   Should I just revert this change?

Can you run another test but add "&conn_only_use_old_forward_open=1" to the tag string?  That should prevent the library from using connected mode and it will just send the packets via the UCMM.  I think I have enough logging that I will see whether it is choosing 504 or 502 bytes.

kyle-github:
Urg.  Too many simultaneous conversations.  The library will use connected mode, but not use the Extended Forward Open command so the connection packet size will be limited to a 9-bit value.  That should be 502.

ptsOSL:
[log2.txt](https://github.com/user-attachments/files/15916265/log2.txt)

Seems to be 504? 

Also I have just done a test of reading as much data as possible in a single packet. The most I can read is an array of 1988 SINTS. When you add on the CIP header, it is a total of 1994 bytes. 1996 is not possible, an error is always returned. Therefor I would say that the manual is accurate and the PLC reports the wrong value, and the value in the library should be changed to 1994.

kyle-github:
There is a little more overhead than that.   The CIP header is actually 4 bytes.  Plus you have 2 to 4 bytes for the type info.   But, at least with Rockwell, that is not where the definition of "packet start" is.   With a Rockwell PLC, when you get 4002, you actually start earlier in the whole packet than the CIP header.  

When I read a SINT array of 6000 elements, I get 4041 bytes total back (not an even number???).   Backing up 4002 bytes from that puts me at offset 39, which is probably wrong.   I suspect it is 40.

Here's the breakdown of that data for a connected message.

```
00000    70 00 b2 0f 02 00 2c 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 - EIP header 4018 byte payload
00024    00 00 00 00 00 00 - connection config info
00030    02 00 - start of a 2 item CPF header
00032        a1 00 - a connected address item
                 04 00 - address is four bytes long
                     d9 49 1e 17 - connection ID
00040        b1 00 - a connected data item - negotiated packet size starts here
                 9e 0f  - data is 3998 bytes long
00044            01 00 - connection sequence ID (counts in the payload length!)
00046            d2 00 06 00 - CIP header, partial result error 06
00050            c2 00 - data type SINT
00052                 00 00 03 00 00 00 00 00 00 00 00 00  - data
...
04032                 00 00 00 00 00 00 00 00 00 00 - end of the data
```

4018 bytes plus the EIP encapsulation header (24 bytes) is 4042. Rockwell very rarely has odd numbered lengths.

The actual data starts at offset 00050 with the data type.  Working backward from the 4002 packet size, we get that it starts at the connected data item CPF item at offset 40.  So you have 2 bytes for the data item type.  2 bytes for the data item size.  2 bytes for the connection sequence ID.  And then you start the CIP data with the CIP header.  And you have the type info (which is either 2 or 4 bytes, usually).  So you get 4002 - (2 + 2 + 2 + 4) = 3990 before the data type and 3998 after the data type.

Lining up with the connected data item makes a lot of sense.  That is a single part of the packet.  The data item type, length and data are all treated as one unit.

So if you've got 1996 bytes of data for the CIP response and type info.  That's another 6 bytes, so you should get 1990 bytes of data.  You got 1988.  I think you are right in that it uses a 1994-byte connection packet size.

When I read 600 elements of the SINT array using an unconnected message, I get this:

```
00000 6f 00 02 02 02 00 2d 00 00 00 00 00 a5 bd b7 04 00 00 00 00 00 00 00 00 - EIP header 514 byte payload
00024 00 00 00 00 00 00 - config info
00030 02 00 - two item CPF list
00032 00 00 - Null address item
00034.    00 00 - zero length
00036 b2 00 - Unconnected data item
00038     f2 01 - 498 byte payload
00040         d2 00 06 00 - CIP header
00046         c2 00 - data type
00048             00 00 - start of data
...
00528.            00 00 00 00 00 00 00 00 00 00 - end of data
```

So a total of 538 bytes.   Take off the 504-byte "connection" packet size and we are back to offset 34.  That doesn't make sense.  But if we use 502 bytes then we get to offset 36 and that makes sense.  And it looks like we get 490 bytes of data.

Note that one of the learnings I had recently was that ControlNet can only handle 502 bytes of data item of the CPF.  It looks like 502 is the real number.

kyle-github:
I have found more information on how to do fragmented reads/write with Omron PLCs.   There will still be a need to do both request and reply size checks but with the ability to slice things up we should be able to make it easier.

I have an open issue on that #470 and a related one #472.  Both really depend on #471.

Do you want to continue with this PR?  I think it will require a lot of work to "port" it over once the various refactorings are done.   It will probably be a lot easier to do on the refactored base though.

What are your thoughts on this?  I don't want to turn down all this work but I also do not want to make you do it all twice.

ptsOSL:
> There is a little more overhead than that. The CIP header is actually 4 bytes. Plus you have 2 to 4 bytes for the type info. But, at least with Rockwell, that is not where the definition of "packet start" is. With a Rockwell PLC, when you get 4002, you actually start earlier in the whole packet than the CIP header.
> 
> When I read a SINT array of 6000 elements, I get 4041 bytes total back (not an even number???). Backing up 4002 bytes from that puts me at offset 39, which is probably wrong. I suspect it is 40.
> 
> Here's the breakdown of that data for a connected message.
> 
> ```
> 00000    70 00 b2 0f 02 00 2c 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 - EIP header 4018 byte payload
> 00024    00 00 00 00 00 00 - connection config info
> 00030    02 00 - start of a 2 item CPF header
> 00032        a1 00 - a connected address item
>                  04 00 - address is four bytes long
>                      d9 49 1e 17 - connection ID
> 00040        b1 00 - a connected data item - negotiated packet size starts here
>                  9e 0f  - data is 3998 bytes long
> 00044            01 00 - connection sequence ID (counts in the payload length!)
> 00046            d2 00 06 00 - CIP header, partial result error 06
> 00050            c2 00 - data type SINT
> 00052                 00 00 03 00 00 00 00 00 00 00 00 00  - data
> ...
> 04032                 00 00 00 00 00 00 00 00 00 00 - end of the data
> ```
> 
> 4018 bytes plus the EIP encapsulation header (24 bytes) is 4042. Rockwell very rarely has odd numbered lengths.
> 
> The actual data starts at offset 00050 with the data type. Working backward from the 4002 packet size, we get that it starts at the connected data item CPF item at offset 40. So you have 2 bytes for the data item type. 2 bytes for the data item size. 2 bytes for the connection sequence ID. And then you start the CIP data with the CIP header. And you have the type info (which is either 2 or 4 bytes, usually). So you get 4002 - (2 + 2 + 2 + 4) = 3990 before the data type and 3998 after the data type.
> 
> Lining up with the connected data item makes a lot of sense. That is a single part of the packet. The data item type, length and data are all treated as one unit.
> 
> So if you've got 1996 bytes of data for the CIP response and type info. That's another 6 bytes, so you should get 1990 bytes of data. You got 1988. I think you are right in that it uses a 1994-byte connection packet size.
> 
> When I read 600 elements of the SINT array using an unconnected message, I get this:
> 
> ```
> 00000 6f 00 02 02 02 00 2d 00 00 00 00 00 a5 bd b7 04 00 00 00 00 00 00 00 00 - EIP header 514 byte payload
> 00024 00 00 00 00 00 00 - config info
> 00030 02 00 - two item CPF list
> 00032 00 00 - Null address item
> 00034.    00 00 - zero length
> 00036 b2 00 - Unconnected data item
> 00038     f2 01 - 498 byte payload
> 00040         d2 00 06 00 - CIP header
> 00046         c2 00 - data type
> 00048             00 00 - start of data
> ...
> 00528.            00 00 00 00 00 00 00 00 00 00 - end of data
> ```
> 
> So a total of 538 bytes. Take off the 504-byte "connection" packet size and we are back to offset 34. That doesn't make sense. But if we use 502 bytes then we get to offset 36 and that makes sense. And it looks like we get 490 bytes of data.
> 
> Note that one of the learnings I had recently was that ControlNet can only handle 502 bytes of data item of the CPF. It looks like 502 is the real number.

Ahh, yeah it seems the discrepancies we are seeing are due to different interpretations of where the payload should be counted from. Taking another look at my wireshark capture, as you said, the 1994 bytes is 1988 bytes of data + 4 bytes of CIP header + 2 byte data type.

But before this, there are 46 bytes of Ethernet/IP header and connection info. Part of this is 2 bytes which gives the length of the data as CC 07 or 1996. This is because it is including the 2 byte "CIP Sequence Count" as being part of the CIP data, the same as you are seeing for Rockwell. I wouldn't have expected this to be included, but it is. I suppose when deciding what size to define in libplctag, it depends on where you start counting from when you create messages.

My results below look very similar to yours

```
00000    70 00 e0 07 84 01 85 45 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 - EIP header 2016 byte payload
00024    00 00 00 00 00 00 - connection config info
00030    02 00 - start of a 2 item CPF header
00032        a1 00 - a connected address item
                 04 00 - address is four bytes long
                     92 1b e1 10 - connection ID
00040        b1 00 - a connected data item - negotiated packet size starts here
                 cc 07  - data is 1996 bytes long
00044            03 00 - connection sequence ID (counts in the payload length!)
00046            cc 00 00 00 - CIP header
00050            c2 00 - data type SINT
00052                 00 00 00 00 00 00 00 00 00 00 00 00  - data
...
02030                 00 00 00 00 00 00 00 00 00 00 - end of the data
```

ptsOSL:
*Disregard*

If the changes in this branch look reasonable to you, I am happy to recreate them on a branch copied from the latest version of release? Pretty sure it will be quicker than rebasing. I am not super confident with rebasing and not messing up commit messages etc.

I guess I should leave the Omron max message size at 1996? If other PLCs include the extra 2 bytes, then it makes sense to do it for Omron?

ptsOSL:
Disregard that previous message, I hadn't properly understood what you wanted to do. If you are going to implement this "Omron style fragmentation" then it would be more natural to implement the response size checks as part of that process and therefor my changes on this PR wont be useful. Are you planning on making all of these Omron changes yourself? I'm not sure how much time I will have to contribute code, although I can always test things for you.

I think this PR can be closed, but if you decide to abandon the fragmentation or its taking a long time, then you can always reopen this, as it does fix the errors where too much response data is packed together.

kyle-github:
I am under a fairly strong time limit as I leave for 3+ weeks of vacation on Wednesday :-)

I will definitely implement the Omron-style fragmented requests.   They are needed for data which is a single unit but does not fit in one packet.   

Omron has this interesting behavior where you get the entire set of data for whatever you reference.  If the tag is an array of some sort, you get the entire tag.  So "myTag" and "myTag[0]" are very very different even if the element count on the second one is the same as the number of element in the array.  Because of this, you can get a large chunk of data fairly easily.

I think we will still need to have these calculations, but I don't know exactly where they will need to be done and what they will do.  If we have the ability to read/write in chunks, it will be more flexible, but Omron makes us work for our money here.  They do not do it automatically like Rockwell.  

What happens if you read a tag that is bigger than the connection packet size?  Do you just get an error or do you get some partial data?  Rockwell returns some partial data and a status of 0x06 which indicates that we got partial data.

ptsOSL:
Yes, I thought you mentioned you were off on vacation, I hope you have a good time :)

And yes I am well aware of the the differences between myTag and myTag[0] on Omron, although I assumed it was the same for Rockwell. 

If you read a tag bigger than the max packet size you just get an error rather than partial data. I just accepted this as a limitation, but looking at the issue you created, maybe its possible to do this cleanly using the data segment descriptor. Alternatively you could force your user to tell you about the data you are accessing so that you can split up read/requests. i.e if they wanted to get an array that was too big, you could force them to define its datatype and size so that you could manually split it up.

kyle-github:
The data segment descriptor seems pretty flexible.  What I do not know is whether you can ask for more data than is in the tag or not.  I really do not see how we are going to solve all this without getting some type info from the PLC.  That, at least, we are making some progress on.

kyle-github:
So it looks like we can get everything we need to calculate the various sizes.  We can get

- the size in bytes of a UDT tag.
- the size in bytes of an array element.
- the type of an array element (exact when a base type, otherwise we can tell it is a UDT or array).
- the number of dimensions of an array.
- the count of entries in each dimension.

We have some type info. Not a lot. We have it for all the base types.  We do not have the length of the type info for UDTs and arrays.

A long as we leave some reasonable fudge room for the type data, it should be possible to calculate, in advance, how much each tag we can read or write.  That means that we can fill out the data segment/fragment stuff.

I think this is going to be doable!

kyle-github:
OK. I'll leave you with a (hopefully!) working branch `split_omron`.   This has the Omron code duplicated into src/protocols/omron.  It is a copy of what is in the src/protocols/ab folder, but with all the non-Omron stuff removed.  At least I hope I got it all removed.     All the PLC/5, SLC and MicroLogix support is gone for instance.

I was not able to finish with the refactoring that I wanted to do.  I am part way through it.  I am peeling it apart in layers.   So to create a read request:

```c
int build_read_request(omron_tag_p tag, int byte_offset)
{
...
 get a request
...

    uint8_t *cip_start = request->data;
    uint32_t cip_capacity = request->request_capacity;

    rc = conn->eip.reserve_send_request_overhead(&cip_start, &cip_capacity);
    if(rc != PLCTAG_STATUS_OK) {
        .... handle error ....
        return rc;
    }

    rc = conn->cip.encode_read_request(cip_start, &cip_capacity, tag->encoded_tag_name, tag->encoded_tag_name_size, tag->elem_count, tag->offset, amount_to_read);
    ...
    
    rc = conn->eip.encode_send_request(request->data, request->capacity, cip_capacity);

    ....

    /* add the request to the conn's list. */
    rc = conn_add_request(tag->conn, req);

    if (rc != PLCTAG_STATUS_OK) {
        pdebug(DEBUG_ERROR, "Unable to add request to conn! rc=%d", rc);
        tag->req = rc_dec(req);
        return rc;
    }

    /* save the request for later */
    tag->req = req;

    pdebug(DEBUG_INFO, "Done");

    return PLCTAG_STATUS_OK;
}

```

I hope that gives a flavor of what it will be like.  Some key points I am trying to achieve:

- Almost no code branches or differences when dealing with connected vs. unconnected.  This is handled by having a sort of vtable in the session/connection object.  You can see the conn->eip and conn->cip to get to the two different sets of PLC-specific functions.  
- Break out of common elements and functionality into layers.  The EIP header handling above shows this.  First we reserve space, which then sets up the cip payload pointer and buffer size.  Then we create the  read request, which is done generically as CIP doesn't know a thing about connected vs. unconnected.

PR #475, opened 2024-06-26
Modify ab_server to allow tag type BOOL when using CIP-based PLC

jshan0120:
This PR modifies ab_server and there are two big changes.

First of all, it changed to allow tag type BOOL when using CIP-based PLC.
BOOL type takes 0x00C1 to indicate its type and uses size of 1 byte.
I checked this operates well on .NET wrapper.

Secondly, I found that calculation of total_request_size in handle_write_request is wrong.
I think total_request_size should be slice_len(input) - offset - 1, so I changed that.

PR #478, opened 2024-07-16
Improve ab_server

jshan0120:
I found that the server couldn't process bundled requests so I start to improve to use.
Two big features are implemented.

1. The server now receives and responses to bundled requests.
A new tag command `CIP_READ_BUNDLE ` represents bundled read requests, and generate a response packet.

2. The server can accept maximum 2 clients at the same time.
Previously, the server only accept one client.
I need at least two clients at once, so I changed those codes.

In addition to those changes, I added padding value to SINT and BOOL type tags (issue from #475 )

I tested those changes using ControlLogix.
The environment I used was docker and windows OS, with .NET Wrapper.
I think I need more cross-platform tests.


kyle-github:
I just merged this into the improve_server branch.   I will do some testing before I merge it into the prerelease branch.

kyle-github:
I started working on this.   I got a little carried away and now have most of the code rewritten to use threading.   I also changed how the slices worked because there were a number of errors I kept making.   I copied some slice-like ideas from another project I am working on so that changed too.

I hope to have this mostly working in a couple of days.  Unfortunately, I go for some eye surgery tomorrow so I probably will not be able to code tomorrow.

kyle-github:
Still working.   Not sure I'll finish before I leave on my next vacation on Friday.  This vacation is only about a week.

kyle-github:
I did not get it across the finish line.  However, I may have more time to work while I am on vacation.   We are going to a large lake but due to my eye surgery I cannot go in the water.  I'll bring my laptop and hope my wife does not make me sleep outside.

jshan0120:
Thank you for your work! Have a nice vacation :)

PR #496, opened 2024-10-28
ab_server: fix incorrect handling of single-dimension arrays

wegel:
Hi,

I think I've encountered a bug in `ab_server`, specifically in how it handles single-dimension arrays. I know that it's mainly used for the CI (I'm using it within GoogleTest for unit/integration tests to not depend on actual hardware, works great!), but I thought it might still be useful to report bugs.

## Description

When reading from or writing to a single-dimensional array tag (e.g., `DINT_ARRAY:DINT[3]`), the simulator (`ab_server`) seems to incorrectly calculates the element offset. This appears to be due to the `process_tag_segment` function assuming that all tags have three dimensions, which isn't the case for single-dimensional arrays.

## Steps to Reproduce:

1. Create a tag in the server: `DINT_ARRAY:DINT[3]`.
2. Use the following tag paths in your code:
```c
std::string tag_path_0 = "protocol=ab_eip&gateway=127.0.0.1&path=1,0&cpu=LGX&elem_type=DINT&elem_count=1&name=DINT_ARRAY[0]";
std::string tag_path_1 = "protocol=ab_eip&gateway=127.0.0.1&path=1,0&cpu=LGX&elem_type=DINT&elem_count=1&name=DINT_ARRAY[1]";
std::string tag_path_2 = "protocol=ab_eip&gateway=127.0.0.1&path=1,0&cpu=LGX&elem_type=DINT&elem_count=1&name=DINT_ARRAY[2]";
```
3. Attempt to read from or write to these tags using the simulator.
4. Observe that the data read or written does not correspond to the correct array elements.

## Expected vs actual behaviour

The simulator should correctly calculate the element offsets for single-dimensional arrays, allowing proper read/write operations to DINT_ARRAY[0], DINT_ARRAY[1], and DINT_ARRAY[2]. But the simulator actually miscalculates the element offsets due to incorrect handling of array dimensions.

## Cause of the bug

In the `process_tag_segment` function, the element offset is calculated using all three dimensions, regardless of the actual number of dimensions the tag has:

```c
/* Original code that assumes three dimensions */
element_offset = (size_t)(dimensions[0] * ((*tag)->dimensions[1] * (*tag)->dimensions[2]) +
                          dimensions[1] *  (*tag)->dimensions[2] +
                          dimensions[2]);
```

For single-dimensional arrays, `dimensions[1]` and `dimensions[2]` are initialized to 1, which leads to incorrect offset calculations.

## Proposed fix

Modify the element offset calculation in `process_tag_segment` to account for the actual number of dimensions, and make sure `handle_read_request` and `handle_write_request` use the correct calculated byte offset.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,16 @@ jobs:
         echo "shut down server."
         killall ab_server -INT &> /dev/null
 
+    - name: Test Array Notation Read/Write
+      run: |
+        cd ${{ env.DIST }}
+        echo "start up simulator..."
+        ${{ env.DIST }}/ab_server --plc=ControlLogix --path=1,0 --tag=TestBigArray:DINT[5] &
+        sleep 2
+        echo "test array notation read/write"
+        ${{ env.DIST }}/test_array_notation 127.0.0.1 1,0 TestBigArray 5 2000
+        echo "shut down server."
+        killall ab_server -INT &> /dev/null
 
     - name: Upload ZIP artifact
       uses: actions/upload-artifact@v4
@@ -236,6 +246,17 @@ jobs:
         echo "shut down server."
         killall ab_server -INT &> /dev/null
 
+    - name: Test Array Notation Read/Write
+      run: |
+        cd ${{ env.DIST }}
+        echo "start up simulator..."
+        ${{ env.DIST }}/ab_server --plc=ControlLogix --path=1,0 --tag=TestBigArray:DINT[5] &
+        sleep 2
+        echo "test array notation read/write"
+        ${{ env.DIST }}/test_array_notation 127.0.0.1 1,0 TestBigArray 5 2000
+        echo "shut down server."
+        killall ab_server -INT &> /dev/null
+
     - name: Upload ZIP artifact
       uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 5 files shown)


kyle-github:
Huh...  That looks like a bug to me.   I think someone else noted this a while back, but I cannot find the email.   I am running the check now.

wegel:
Checking again, I think the proposed fix is incorrect/incomplete; I'll report later, stay tuned. Thanks!

wegel:
I had a few cases where the proposed fix didn't work; I've now amended it, and now all my tests pass. Sorry about that.

kyle-github:
Let me do some checks.   Multi-dimensional arrays are treated as row-major (if I am using the terminology correctly).   So what you get is:

for n[x,y,z] you have:

index_n = x + (y * Xsize) + (z * (Xsize * Ysize))

The way that Rockwell implemented the protocol is such that if an array has N dimensions you must have N indexes or none at all in the tag path.  I.e. "myArray[42,43,44]" or "myArray".  The latter is the same as "myArray[0,0,0]".

Also, strangely, it appears that the array dimensions are reported in reverse order in the raw tag listing data.   If you have a tag with dimensions 10,4 (x and y) then they are reported in the tag listing data as 4, 10, 0.   That might be moot here, but I probably need to fix the tag listing code.

kyle-github:
BTW I moved this merge to the prerelease branch.  I do not add code directly to the release branch.  I really should document that somewhere...

wegel:
Understood; will target the prerelease branch for any future PR. Don't hesitate if you need me to add, modify or test something. Thanks!

wegel:
Added a test for the array notation, which fails without the proposed fix.

kyle-github:
This is failing in my GitHub CI.  I am going to do some testing against a real PLC to make absolutely sure I am doing the right thing.

kyle-github:
Ooooh, boy.   Just opened a can of worms.   The array indexes are used in reverse order.   If you have [x,y,z] the z coord is used as the lowest order coordinate.  I.e. what we usually call "x".   I need to do far more testing here.  This is against a real PLC as I want to make sure that the tag simulator behaves the same.

kyle-github:
I think the simulator is doing the calculation for multi-dimensional arrays incorrectly.  If I have an array tag, MyTag[2,3], the elements, in order are:

[0,0]
[0,1]
[0,2]
[1,0]
[1,1]
[1,2]

Notice that the _last_ index is the one to add and the first one's increment has to be calculated by multiplying by the size of the second dimension.

PR #503, opened 2024-11-23
Multithread ab_server

### LINKED ISSUE #499: ab_server only accepts one connection at a time, so can't be used for PLC simulation
As a new libplctag user I was excited to find `ab_server` in solid use as a CI test tool, and that it has been suggested as a starting point for PLC simulation.

I've created a wrapper that launches `ab_server` with a set of tags, and then calls `tag_rw2` with `--write` every now and then to "play back" a scenario from a file. It works brilliantly, and I can see in the logs and with manual `tag_rw2` read calls that the server represents the scenario state at each point in time. A virtual PLC, presenting pre-meditated data streams, running on the PC!

Unfortunately, as soon as I launch my libplctag-enabled application and have it connect to the simulated PLC, the next `tag_rw2` call to update the scenario data hangs.

After some investigation, I see that libplctag works by opening a connection, queueing up requests, handling a transaction, queueing up requests, handling a transaction, and so on. Only after [5 seconds](https://groups.google.com/g/libplctag/c/g0Zwr0Kxwfc/m/xr43A0ZnBwAJ) will it opt to close the connection between requests.

Fair enough. But alas, `ab_server` only [handles one connection](https://github.com/libplctag/libplctag/blob/release/src/tests/ab_server/src/tcp_server.c#L78) at a time. So my dreams of a PLC simulator seem shot.

The only workaround I can find is to slow my application's polling rate to slower than once ever 5 seconds. Any other suggestions? As others have said, I don't want to burden ab_server's role in life with the requirements of a simulation tool, but it seems so very close!

hraftery:
Resolves #499.

Design notes:

- tx/rx buffer moved from `main()` stack to thread function stack, so each thread has one.
- threads are created on each successful `accept`. The function contents are primarily just what was in the main `tcp_server_start` loop, with variable references updated to suit. I found a couple of wee (mainly debugging) bugs along the way, but copied them over too so those changes could be part of their own commit. That way, the commit that introduces multi-threading should run functionally the same as the single-thread version used to.
- passed around the size of the plc struct, so a copy could be made for each thread. This should be made unnecessary, but for now there's a few connection-specific fields (like `session_handle`), so it was safest to give every thread its own copy. We can still access the shared context (like the tags) via their pointers.
- added a really tight mutex around the data field of each tag. Made the big simplifying assumption that no other shared data is changed after initialisation. So we don't need to protect the list of tags, their names or types, or indeed any other context configuration.
  - note the mutex is never even destroyed, since the tags are never destroyed.
- added `memory`, `mutex` and `thread` modules, which are simply amalgamations of the relevant sections of the `platform/posix` and `platform/windows` modules, in the style of the existing `socket` module. As in `socket`, all that changed was to redefine the `STATUS` and `ERR` defines, change all the `pdebug` calls to `info` and `error`, and to support both platforms by branching with `#if IS_WINDOWS` where necessary. Quite minimal effort really.
  - I have not tested, or even compiled, the Windows version.
  - Stunningly, I found after doing all this that the modules weren't even required at link time. Turns out the `CMakeLists.txt` file adds `${example_LIBRARIES}` to the link command, which adds `tool_lib` which is `plctag_dyn`. So `ab_server` has all the symbols from `platform` available anyway! However, I wasn't able to just `#include <platform.h>` (and `#include <lib/libplctag.h>`), which was actually my first instinct, because the existing `socket` module redefines some symbols differently. So I went ahead and created the extra modules, and proved that it could be linked with and without `libplctag`. I suspect linking *without* is the desired behaviour, but will leave that for another day.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_ubuntu_x64"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_ubuntu_x64"
       BUILD: "${{ github.workspace }}/build"
       DIST: "${{ github.workspace }}/build/bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
@@ -130,8 +130,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_ubuntu_x86"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_ubuntu_x86"
       BUILD: "${{ github.workspace }}/build"
       DIST: "${{ github.workspace }}/build/bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
@@ -247,8 +247,8 @@ jobs:
     runs-on: macos-latest
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_macos_x64"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_macos_x64"
       BUILD: "${{ github.workspace }}/build"
       DIST: "${{ github.workspace }}/build/bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
@@ -361,8 +361,8 @@ jobs:
     runs-on: macos-latest
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_macos_aarch64"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_macos_aarch64"
       BUILD: "${{ github.workspace }}/build"
       DIST: "${{ github.workspace }}/build/bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
@@ -396,8 +396,8 @@ jobs:
     runs-on: windows-2019
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_windows_x64"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_windows_x64"
       BUILD: "${{ github.workspace }}\\build"
       DIST: "${{ github.workspace }}\\build\\bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
@@ -518,8 +518,8 @@ jobs:
     runs-on: windows-2019
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_windows_x86"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_windows_x86"
       BUILD: "${{ github.workspace }}/build"
       DIST: "${{ github.workspace }}/build/bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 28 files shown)


kyle-github:
This is not compiling on some of my CI machines.  Specifically Windows.   I will try to fix it.

hraftery:
Ah yes. It could be as simple as adding `#include <windows.h>` (and others) to the `#if IS_WINDOWS` include sections at the top of `mutex.c` and `thread.c`. I don't know sorry, I'm not set up to test on Windows at the moment and only guessed it'd be included transitively.

kyle-github:
I am still working on some other code for checks and handling of indexes.   It is far trickier than I thought so I probably missed an easy way to do it.

hraftery:
I feel a little obliged here, since I just realised you'd already started a multithreading implementation on a different branch! Good news is that we appear to have pretty much the same approach, though I benefitted from not trying to refactor `slice` at the same time.

So I enabled workflows on my fork and got stuck into the build issues. Sure enough it was just a missing `#include <windows.h>` or two, which I found just by trial and error, pushing to my `prerelease` branch and reading the CI output.

But there was also something pesky happening with the merge of 09438bb9965fda3635ee2135c951d3d0e0e9dffb. Seems there might have been a simultaneous push or something, and the two commits overwrote each other in two places. I went a bit cross-eyed trying to unroll the merge, so abandoned that and just made the correction in a new commit.

Not knowing quite how to contribute the work, I decided to apply it back to my original `multithread-ab_server` branch, and create a new PR, #504. HTH!

kyle-github:
Thanks!   I am still working on fixing up the index handling.   Once I am done with that I will get this all working and tested and then probably do a release.

kyle-github:
I wrote a tool to test the performance and it is not what I expected.   The tool will be in the examples directory and is called test_emulator_performance.c.

Here is what I see:

| thread count | requests per millisecond |
|----          |----
| 1   | 19 |
|10  | 75 |
| 20 | 78 |
|30  | 76  |
| 40 |  74 |
| 50 | 70 |
| 60 | 71 |
|70 | 73 |

There is quite a bit of variation but the overall trend is pretty consistent.   We max out performance at around 20 threads.

When running these tests the CPU load for the ab_server program is pretty low.   This suggests that it is worth trying a single threaded approach again but with multiple sockets handled all at once.   I will release with this code as is because it definitely helps!  The response times are very low even at the larger numbers of threads.

For this test, each thread reads one tag and each thread uses its own connection group forcing a new connection to ab_server.  Each test runs for 5 seconds.

I ran the ab_server program under Valgrind and it was not particularly happy.  I will not delay the release for this, but there are definitely some leaks:

```
==174367== HEAP SUMMARY:
==174367==     in use at exit: 9,161 bytes in 311 blocks
==174367==   total heap usage: 1,137 allocs, 826 frees, 121,881 bytes allocated
==174367== 
==174367== 4,313 (96 direct, 4,217 indirect) bytes in 1 blocks are definitely lost in loss record 5 of 6
==174367==    at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==174367==    by 0x111D32: parse_cip_tag (main.c:544)
==174367==    by 0x110FF7: process_args (main.c:326)
==174367==    by 0x110883: main (main.c:154)
==174367== 
==174367== 4,848 bytes in 303 blocks are definitely lost in loss record 6 of 6
==174367==    at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==174367==    by 0x112AF1: mem_alloc (memory.c:55)
==174367==    by 0x11674B: thread_create (thread.c:82)
==174367==    by 0x1160AD: tcp_server_start (tcp_server.c:117)
==174367==    by 0x1108DF: main (main.c:159)
==174367== 
==174367== LEAK SUMMARY:
==174367==    definitely lost: 4,944 bytes in 304 blocks
==174367==    indirectly lost: 4,217 bytes in 7 blocks
==174367==      possibly lost: 0 bytes in 0 blocks
==174367==    still reachable: 0 bytes in 0 blocks
==174367==         suppressed: 0 bytes in 0 blocks
```

Note that this might be because I killed the process with ^C.  I will add a signal handler and see if that helps after I release this.

Thanks for your work on this!

hraftery:
Oh yes! `thread_destroy()` needs to be called on `session->thread`. That'll definitely leak otherwise. I missed that. And it's a bit tricky because the thread free's `session` when it ends, and the parent currently doesn't keep a reference to `thread`. Nor does it know when the child dies.

The performance stats are interesting. I can imagine running into various other limits depending on the environment. I might take a look at that test at some point when it's available.

PR #504, opened 2024-11-27
Fix build issues with PR #503

hraftery:
Two simple fixes to pass CI:

1. Correct very small conflict in the merge of "Multithread ab_server" and "ab_server: fix incorrect handling of single-dimension arrays".
2. Fix Windows build by including a couple of general windows headers.

Despite the simplicity of these changes, I couldn't untangle it in the history so have made the corrections as new commits instead and created a new PR from the same branch.

PR #511, opened 2025-01-08
Fix #510 (miss-matched tag names)

### LINKED ISSUE #510: ab_server mixes up tags that have the same prefix
Issue title is yet to be proven, but is the current theory, after noticing that writing to a tag named `WORD1_WORD2_WORD3_WORD4_ACK` seems to overwrite the value in tag named `WORD1_WORD2_WORD3_WORD4`.

The issue seems to be here:

https://github.com/libplctag/libplctag/blob/release/src/tests/ab_server/src/slice.h#L58

which is called here:

https://github.com/libplctag/libplctag/blob/release/src/tests/ab_server/src/cip.c#L643

To wit, tag names are matched by comparing the strings, but only up to the number of characters in the existing tag?

---

There are several other confounding factors that I hope prove irrelevant once I investigate the smoking gun above:

1. I'm writing from one client and reading from another (using my multi-threaded fork).
2. Quickly (read/writes occur every 50ms or so, asynchronously).
3. The `ACK` tag does not exist on the server - I'm expecting the write operation to just get dropped.

hraftery:
Make `ab_server` lookup tag names based on their entire name, not just their prefix. See single commit message for details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_ubuntu_x64"
+      VERSION: "2.6.3"
+      ARTIFACT: "libplctag_2.6.3_ubuntu_x64"
       BUILD: "${{ github.workspace }}/build"
       DIST: "${{ github.workspace }}/build/bin_dist"
       BRANCH: ${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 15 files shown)


kyle-github:
I moved this to the prerelease branch.   That's where all the work is happening right now.

hraftery:
> I moved this to the prerelease branch. That's where all the work is happening right now.

Cool. I think the tests are failing because the wrong ref is configured:

https://github.com/libplctag/libplctag/blob/prerelease/.github/workflows/ci.yml#L28

Looks like the ref setting needs to be removed, so the default can take effect. The [default](https://github.com/actions/checkout?tab=readme-ov-file#usage) knows it's a PR, so not to go looking for the source by the branch name. The default seems to be set [here](https://github.com/actions/checkout/blob/main/src/input-helper.ts#L63) but I'm out of my depth.

Note if there are other test issues, I fixed some recently here: https://github.com/libplctag/libplctag/pull/504

Sorry if I'm confusing things all over the place here, I'm a bit lost following all the merges and may not be contributing well!

Lastly, the issue didn't seem to auto link to this PR, so tagging it here: #510

kyle-github:
I think I have this fixed now.   I did more brain surgery than you had.  I ended up splitting some of the slice functions into exact and partial matches.

kyle-github:
Note that it does not build on Windows yet.   Coming soon.

kyle-github:
It builds on Windows.   I am going to skip this PR for now.   I split the match functions in slice.h to have exact and prefix matching.  I use only the exact version in the tag name matches.   I will have a release out within a few days.   Thank you very much for noting this problem!!!!

Issue #515, opened 2025-01-20
Memory leak on connection loss / PLCTAG_ERR_TIMEOUT


HalfenD:
## Description
When reading a tag and a disconnect/**PLCTAG_ERR_TIMEOUT** occurs, there is a memory leak.  
I analyzed the error with `heaptrack` and found this line allocated the memory that leaked: https://github.com/libplctag/libplctag/blob/c55bc5876d938dda1c609750cde5ae4812d7b8a8/src/protocols/ab/session.c#L2778

## Steps to Reproduce:

Create a tag with for example with auto_sync_read. (manual read results in the same)
Disconnect the network between some request and now.
Observe the memory leak.

## Expected Behavior:
There should be no memory leak when a disconnect occurs during a read process.

## Environment:
Software Version: V2.6.3 and release-branch
PLC: ControlLogix EIP_1769
Operating System: Debian-Linux amd64

kyle-github:
I think I need to spend more time looking for more of these.

kyle-github:
This should be fixed with 2.6.4 but I need better tests.   If you see this again, please reopen this or open a new issue.  Again, thank you for finding and filing this!

PR #516, opened 2025-01-21
Fix memory leak after request timeout (#515)

HalfenD:
After a short analysis, I found that `tag->req` was set to **NULL** without freeing the actual request object. After applying this fix in my test, the memory leak did not occur anymore. Although I'm not very familiar with the library code, I believe this is the correct approach.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      VERSION: "2.6.2"
-      ARTIFACT: "libplctag_2.6.2_ubuntu_x64"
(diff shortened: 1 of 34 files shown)


kyle-github:
This looks good.  I usually do something like 

```c
tag->req = rc_dec(tag->req);
```

That way it sets it to NULL locally to make sure that if we call rc_dec() again on it accidentally, we'll get a warning but no dereferencing of NULL or other fatal problems.

I'll pull this in tomorrow once I get the build working again on Windows.

kyle-github:
I'll start working on this now.  I have it all compiling in Windows.  There is a possible double free as well that I need to work on.

kyle-github:
All merged!

PR #530, opened 2025-05-06
auto_sync_read_ms failing on reconnects

wegel:
I seem to have found a bug where the `auto_sync_read_ms` setting stops working after a disconnect/reconnect cycle. The PR adds a reproducer test and shows how to *potentially* fix the issue.

I say "seem", because I'm not quite sure this is a bug vs intended, as I find it suspicious that such a "big" bug has gone unnoticed for that long. But this basically popped up within 5 minutes of testing the lib with an actual PLC (ControlLogix), when using `auto_sync_read_ms` tags.

What seems to happen is when a tag with auto_sync_read_ms enabled gets disconnected (e.g., network failure) and then reconnects, the automatic reading stops working. This is because the auto-sync timer state isn't properly restored during reconnection.

The proposed fix ensures that the auto_sync_read_ms setting is properly preserved when a tag reconnects. This PR also includes a test case that demonstrates the issue and verifies the fix.

Additionally, I've fixed the CMake include paths to properly support using libplctag as a subproject via eg FetchContent (basically you can never use `CMAKE_SOURCE_DIR` in your CMakeLists.txt, as that refers to the root directory of the "main" project, which might not be libplctag).

It's targeting `release` as I'm currently using/testing v2.6.5 so my internal fix was based off that, and it doesn't apply without conflict on `prerelease`.


kyle-github:
Thanks!   I thought I got all the CMAKE_SOURCE_DIRs left.   Grrr.

Let me look at the auto sync stuff.  It is finicky and fragile.  When I start doing more refactoring to eliminate duplicate code, it is one of the first parts I am going to redo.  That and the event handling for the callback function.

kyle-github:
Looks like there might be a GitHub problem.  Git is failing to check out the code so all the CI jobs are failing.  I'll retrigger later today to see if that is the issue.

kyle-github:
So... the auto stuff is in a gray area.  If a PLC goes offline do you want all outstanding requests to retry or not?  I may add an additional attribute to control this and an event type to make sure it gets to the user.

However, you have found a bug in that there is a race condition here. If the request is still in the queue and has not been sent, then it will go through when the PLC reconnects.   But if the request has been sent then it will not retry.  So it is inconsistent depending on the timing.  That is definitely a bug!

As to why it has not been caught, I probably do not have a test for it.  I do have tests that test reconnect, but I will need to check if any of them use auto sync.  I will take a look at your test and also see if there is anything I can do with the ab_server to make this part of the CI checks.

wegel:
What I did for my test was to make it manage its own `ab_server`, so it can kill it and then restart it to simulate a hard reset.

> If a PLC goes offline do you want all outstanding requests to retry or not?

Personally, for non-auto-sync requests, no; if there's any kind of error (including a timeout due to a network error or similar), I want to be notified and then I'll decide what to do. AFAIK, that's already how it works.

For auto-sync requests, if what you mean by `do you want all outstanding requests to retry` would be that for an auto-sync tag set to 100ms, and the PLC is offline somehow for 10s, when it comes back we'd receive 100 callbacks, then no; those auto-sync operations that occur during network/PLC downtime are effectively lost, and that seems correct to me, especially if we're notified of that in the callback (currently I'm raising `PLCTAG_EVENT_ABORTED` and setting `PLCTAG_ERR_TIMEOUT`, but not sure if that's appropriate). But obviously when/if the network/PLC gets back online, we want auto-sync to still work if we haven't explicitly removed it.

Makes sense?

kyle-github:
You make a good point.  I think that the retry is going to be a bit tricky because the actual status of the PLC connection is not available to the tag without some internal trickery.  I think the reset to retry should happen on the PLC thread-side.   I will look into that this weekend.  

Note that for non-auto sync tags, you always have a timeout whether that is passed into the API or enforced by the application.   But for auto sync tags that timeout isn't meaningful because it is internal and there is already logic to skip triggering new reads when more than one read period has elapsed.  Hmm... that actually makes some sense.  The auto sync read period is implicitly a timeout.  Let me run with that for a bit.

wegel:
Yep, I think that what you're describing is pretty much what I did here: https://github.com/libplctag/libplctag/pull/530/commits/637363fe3ebc94174e788c9388b8a98fd4f0042f#diff-95f6d371f9f807b6e677e522e76145563118e420c5e1d5e0f66dc8c145e891d6R393

kyle-github:
Ooops.  Didn't mean to close this.  I was cleaning up branches and managed to kill this off.  I think I can revive it.

kyle-github:
Reopening.

kyle-github:
Something is weird here.  It keeps failing all the CI checks on a git failure.   I'll create a branch for this and try to merge it in.

kyle-github:
All is merged on a new branch `merge_in_auto_sync_fixes`.  However, the test program is failing on my Apple laptop.  The auto sync tag does not continue to work after the server is bounced.  Can you check out that branch and see what I have messed up?   I've gone through the changes a couple of times and it looks like it is correct to me.

BTW, I also ported the test code to work on Windows.  I have not tested it yet, but it is not going to be too hard to make it work now.

kyle-github:
Digging into this is going to be my focus for the next day or so.

wegel:
Sorry, haven't had the chance to retest it in the last few days; I'll let you know if I can test it, but probably won't be today. I'm on linux by the way, haven't tried the test program elsewhere yet. Thanks!

kyle-github:
I think I see what is happening.  The underlying IO thread for the PLC connection is not reconnecting before the test finishes.   That may be due to the changes I made in some of the sleep calls.   But it still should work.   I am not outputting the time that the IO thread is waiting to try a reconnect so I am not sure what it is doing.

kyle-github:
I am trying to add some simple back off on retries starting at 100ms and doubling on each failure until 10s.  Will report back if this works.   The logic in the state machine is already annoyingly complicated.   Must do something about that soon.

kyle-github:
I have been experimenting to see how to solve the problem of failure to read after a disconnect.  The underlying socket is reopened but the tag read is not restarted in some circumstances (depends on timing and some other factors).   

I need to solve this a different way.   Instead of having the "client side" do a timeout, I will have the lower levels push the in-flight requests back on the queue.  Right now any inflight request that is left when the connection drops is terminated.   There should be a handshake there but there isn't.  I really need another error code but I am not excited about creating a new one as that is essentially an API change.

Issue #498, opened 2024-11-12
ab_server: issue with multithreaded writes


wegel:
Hi,

   I seem to have stumbled on a bug in ab_server for multithreaded writes. Again, I know that ab_server is only an internal tool, but there's so few small and (relatively) simple PLC simulator, I think it's worth it trying to fix it.

   This test fails: [test_multithreaded_write.c](https://github.com/wegel/libplctag/blob/ab_server-multithreaded-write/src/examples/test_multithreaded_write.c) (see [the CI run here](https://github.com/wegel/libplctag/actions/runs/11789713562/job/32838931228#step:17:18)). It's important to note that the test passes when testing with an actual PLC (well, it passes with the one I have on hand, anyway).

   This one passes: [test_sequential_write.c](https://github.com/wegel/libplctag/blob/ab_server-multithreaded-write/src/examples/test_sequential_write.c)

   I'm pretty sure the problem also happens on reads; eg, it seems to be a general thread-safety issue.

   Haven't had a chance to look at the cause yet, if I can find some time, I'll post my findings here.

hraftery:
I _think_ the problem here is that `ab_server` is not multi-threaded?

I think `test_multithreaded_write` is your creation? The failure occurs when you call `plc_tag_create()`, which is long and complex, and I wasn't actually able to find where/why it returns `PLCTAG_ERR_UNSUPPORTED`. But given that you specify a [non-zero timeout](https://github.com/libplctag/libplctag/wiki/API#creating-a-tag-handle), it's likely it's performing a read before returning. _ab_server_ will only accept one active connection at a time, so multiple concurrent reads will be ignored.

I don't know why all 5 threads fail though, so I could well be wrong here.

On the multi-threading issue, I agree `ab_server` is very close to the PLC simulator we desire, so I made it [multi-threaded](https://github.com/libplctag/libplctag/pull/503). Maybe you could try your test on that fork, prior to its release?

kyle-github:
I believe the underlying problem was the single-threaded nature of ab_server.  It is now multi-threaded, so please tests 2.6.4 Beta!

PR #548, opened 2025-07-03
Simultaneous Read/Write on a Tag using the ab_server issue

dominoplayer84:
This PR introduces a test that reproduces an issue occurring during simultaneous read and write operations on the same tag across two concurrent threads when using ab_server.

The test is expected to fail currently, as no fix has been applied yet. Its purpose is to document and reliably reproduce the issue, so that a proper solution can be developed and verified against it.

**Details**:

- Two threads access the same tag concurrently: one performs writes while the other performs reads.
- This setup highlights potential race conditions or data consistency problems under concurrent access.

Please note that this is not a fix — only a failing test to assist in diagnosing and resolving the problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,17 +275,6 @@ jobs:
         echo "shut down server."
         killall ab_server -INT &> /dev/null
 
-    - name: Test Array Notation Read/Write
-      run: |
-        cd ${{ env.DIST }}
-        echo "start up simulator..."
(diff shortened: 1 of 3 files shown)


dominoplayer84:
Related  to #498

kyle-github:
Thanks!  I have been meaning to do a serious rewrite of the ab_server code.  It was literally something I hacked up on vacation when I had no Internet access.  It got polished a little since then but never really cleaned up.

kyle-github:
Thanks for this! I think I am in a place where I can fix this now.

Issue #572, opened 2025-08-19
Can not write individual bits to micrologix 1400


mikecflood:
I think this is the same issue as #565 but for the micrologix 1400. Now that issue https://github.com/libplctag/libplctag/issues/570 has been fixed I was able to create and read tags for the micrologic 1400. However I was not able to write individual bits.

The tag I was using is:

protocol=ab_eip&plc=micrologix&gateway=[DNS_NAME]&name=B9:0/0&elem_size=2&elem_count=1

This is the level 4 debug: [debug_level_4.txt](https://github.com/user-attachments/files/21860054/debug_level_4.txt)

If it helps I was able to write an int, this is the tag used to write to an int which worked:

protocol=ab_eip&plc=micrologix&gateway=[DNS_NAME]&name=N11:0&elem_size=2&elem_count=1

kyle-github:
Sorry for the delay.  This one confused me for a long time.  The problem seems to be that you have a "elem_size" attribute.  There is some interaction here that I need to trace down.  Here is what I get when I leave that out (this is against a real Micrologix):

```
./tag_rw2 --type=bit '--tag=protocol=ab-eip&gateway=10.206.1.36&plc=micrologix&elem_count=4&name=B3:0/6' --write=1
Library version 2.6.10.
Processing argument 1 "--type=bit".
Processing argument 2 "--tag=protocol=ab-eip&gateway=10.206.1.36&plc=micrologix&elem_count=4&name=B3:0/6".
Processing argument 3 "--write=1".
Number of write args 1.
bit 6 = 0
New values written to tag.
bit 6 = 1
```

The `elem_count` attribute is ignored, but it looks like `elem_size` is not ignored, or is only partially ignored.

Trying leaving the `elem_size` attribute out of your tag strings and see if that works.

I am going to call this a bug for now because a bit tag should ignore the element size for a PCCC/DF1 PLC.

Thanks for bringing this up!

kyle-github:
Hmm, nope.  I think you uncovered another bug too.  The mask needs to be different for SLC/Micrologix and PLC/5.  My DF1 docs are... unclear here.

kyle-github:
Yep, another bug.   It turns out that the mask for a Micrologix/SLC when writing a data file element is two bytes. Not more.  Not less.  The command being used is not really documented.  I have a PDF of a photocopy of what looks like an errata sheet that has handwritten notes that is my "documentation" for this.

[...]

PR #576, opened 2025-09-04
v2.6.11 - More fixes for Micrologix/SLC bit writes, fix issue #572

kyle-github:
Thanks again to user @mikecflood for finding this.   Writes to single bits of SLC or Micrologix were not working correctly.   Unfortunately, the functions supported by the SLC/Micrologix PLCs do not allow for good RMW (read-modify-write) support.   Oddly, the PLC5 series has much better support!

On a SLC/Micrologix PLC you can only write with a 16-bit mask.   Due to that I have limited the support for single bit writes (via bit tags) to those tags that are 16 bits.  This includes single elements from the B and N files.  You can read a bit tag from nearly anything, but not write.

I would really appreciate it if those of you who have SLC or Micrologix PLCs hung off of DH+ that can be bridged from a ControlLogix could test this.   I do not have such a configuration so the code for that path is untested.

This fixes issue #572.

You can find this release [here](https://github.com/libplctag/libplctag/releases/tag/v2.6.11).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -539,7 +539,7 @@ jobs:
 
   windows_x64_MSVC:
 
-    runs-on: windows-latest
+    runs-on: windows-2022
 
     env:
       VERSION: "2.6.11"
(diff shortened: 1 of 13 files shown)


Vadoola:
I have a customer site that still has a few SLCs on DH+ that we typically access through a ControlLogix Bridge. I'll see if I can write up a quick test, and give it a go on their next down day, but I can't guarantee if I'll get a chance to properly test it or not.

PR #581, opened 2025-09-20
Build for Alpine/musl.

kyle-github:
Adding build process for Alpine using musl instead of glibc.


timyhac:
Hey - not sure if its useful - there a pre-built toolchains [available from bootlin.](https://toolchains.bootlin.com/downloads/releases/toolchains/armv5-eabi/tarballs/)

PR #591, opened 2025-11-08
Start adding in reactor/buf/fsm code.  

kyle-github:
First step toward refactoring the main tag handling.  This updates the modbus server code and test scripts.


kyle-github:
Hello @mikecflood, if you have a chance, please try the new Modbus code with the v2.6.13-pre prerelease.  There are a lot of changes and I do not have real Modbus devices on which to test.  I rewrote the Modbus server and tuned it and I now get about 150-200us (microseconds) for a request to be processed and the response to be sent back.   So I can increase the testing pressure on my client Modbus code.

mikecflood:
I rebuilt my program with this version and ran my regression tests, which sadly failed while testing the COMPACT_LOGIX. So I have not yet tested the modbus part.

As it is just from my tests I am including the level 4 debug messaging for the entire test. It seems the first two test ran fine but when the test tried to create a tag for the third test it failed (odd because it creates the same tag for each test).

There is a mix of messages from you lib and mine but the two are distinctly different so it should not be hard to tell whos message is whos.

[libplctag_test_output.txt](https://github.com/user-attachments/files/23894614/libplctag_test_output.txt)

PR #592, opened 2025-11-27
Make windows imports mingw compatible

n3-ruben-sanchez:
Under Linux mingw is case sensitive, if we want to cross compile capital W breaks compilation.

### DIFF
--- a/src/examples/compat_utils.h
+++ b/src/examples/compat_utils.h
@@ -66,7 +66,7 @@ typedef pthread_cond_t compat_cond_t;
 
 #    define WIN32_LEAN_AND_MEAN
 
-#    include <Windows.h>
+#    include <windows.h>
 
 #    include <time.h>
 
--- a/src/platform/windows/platform.c
+++ b/src/platform/windows/platform.c
@@ -38,11 +38,11 @@
 #include <platform.h>
 
 /* KEEP THE SPACES BETWEEN THE INCLUDES.  The order is required! */
-#include <Winsock2.h>
+#include <winsock2.h>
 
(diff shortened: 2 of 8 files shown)


kyle-github:
Thanks for these PRs!  I never tried to use MinGW on Linux, but it makes sense that it would be case sensitive.  Do you know if there is anywhere to find a definitive list of all the data types?  I have not found one so I have to depend on other projects like Pylogix or Pycomm3 to find them out or from users of the library like you!

There is a problem with GitHub right now and the CI tests are failing, so I will try to get these included into the prerelease branch in the next day or two.  It is a holiday where I am today.

n3-ruben-sanchez:
Wow, fast feedback here!

Thank you for this great library and the maintenance.
I'm using Nix to build x86_64-linux, aarch64-linux and x86_64-windows via mingw64 from x86_64 host.
And I thought probably someone else can need it too and is not a big change.

> Do you know if there is anywhere to find a definitive list of all the data types?

I don't, I'm using the FactoryTalk Echo as ControlLogix simulator and building downloading programs with the new types I need to support. If you want more info about this feel free to ask!

> There is a problem with GitHub right now and the CI tests are failing, so I will try to get these included into the prerelease branch in the next day or two. It is a holiday where I am today.

No worries, enjoy your free time!

kyle-github:
Your PR is merged into the prerelease branch.  I had to do some manual fix ups because I updated the prerelease branch a lot and the patch did not apply cleanly.  Thanks!

If you want to try it, check out the prerelease branch.

PR #605, opened 2025-12-06
There was a possibility of race in plc_tag_abort() 

kyle-github:
A direct call to plc_tag_abort() could be delayed such that one of the handler/tickler threads started first and aborted any ongoing operation.  The race condition is that if a totally different thread called plc_tag_read() and started a read and then the thread that called plc_tag_abort() got into the tag mutex, it would unconditionally clear the read_in_flight and read_complete flags.  So a read or write could be lost.


kyle-github:
This was merged from another branch.

PR #618, opened 2026-03-01
ab_server: scope CIP_SRV_UNCONNECTED_SEND case body

chenrui333:
This scopes the `CIP_SRV_UNCONNECTED_SEND` case body in `src/tools/ab_server/cip.c` so declarations are valid in C11 builds.

Related Homebrew source-build failure and fix context:
- https://github.com/Homebrew/homebrew-core/pull/269933

### DIFF
--- a/src/tools/ab_server/cip.c
+++ b/src/tools/ab_server/cip.c
@@ -168,7 +168,7 @@ slice_s cip_dispatch_unconnected_request(slice_s input, slice_s output, plc_s *p
 
         case CIP_SRV_PCCC_EXECUTE: return dispatch_pccc_request(input, output, plc); break;
 
-        case CIP_SRV_UNCONNECTED_SEND:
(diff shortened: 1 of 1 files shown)


kyle-github:
Ugh.  I had that already patched but I managed to miss it when building the release.  It is already in the next release updates.  I moved the definition of to the top of the function instead of adding braces.

Can you live with this patch on your side until the next release?  I try to release fixes when they are needed but I don't want to overwhelm downstream packagers either.

kyle-github:
I checked in the fix in the prerelease branch.  Can you check to make sure that it works for your build?

Do you use other flags for building for Homebrew?  If so, can you tell me what they are so that I can make sure I replicate them to avoid this in the future?

kyle-github:
Thanks again for pointing this out!  Please let me know if you are still getting warnings.

Issue #631, opened 2026-06-08
TCP connection not gracefully closing (noticed with modbus)


mikecflood:
Note: while I am the one using this library, one of my co-workers that handles the networking noticed this and brought it to my attention, so please forgive me if this is poorly described or missing important information.

This issue was noticed on modbus, I don't know if it applies to the other protocols.

When a connection is closed (all tags destroyed) a TCP packet is sent with the RST flag set. I am told an orderly shutdown consists of an exchange of 4 packets with FIN and ACK flags, where the first peer signals to the other that it wants to terminate the connection. The other peer acknowledges this and signals its own intent to terminate the connection. The first peer then acknowledges this. Then the connection is considered properly closed.

On linux, which is what I am using, the above is performed by the TCP stack of the operating system automatically when an application closes a network file descriptor, including the case when the application calls "close()" on a socket. However, there are conditions where this does not occur and an abrupt termination of the connection with the RST flag is sent. The two most likely causes of an abrupt termination are calling "close()" while there is unread data in the TCP buffer associated with the connection or while a request is mid-flight (after sending a read request, but before receiving a reply).

kyle-github:
Thanks for bringing this up.  I will look into it.  Both Modbus and EtherNet/IP use the same socket library internally so it may be at that level.  I'm going to call this a bug since it is not polite to shutdown a connection like that :-)

mikecflood:
Thanks for your work and looking into this.

There is one other thing my co-work brought up that he noticed while monitoring the network traffic.

Here is the scenario: there are two tag that exist for one PLC (in this case modbus), a plc_tag_read() is called for each. Immediately after that plc_tag_destory() is called on each in sequence. At this point the first read request is already mid, now before getting a reply the second read request is sent. A reply to each is sent back from the client and libplctag sends back two RST packets to the client.

From what I understand it normally waits for a reply to the first request before sending the second request.

Theory: If the second request is being sent because plc_tag_destory() was called on the first this would indicate that if a tag is destroyed while a read/write is mid-flight it would be dropped and the next request sent without waiting, and it might collide if the connection was not being closed.

mikecflood:
Possible fix:

I don't normal program directly with sockets so this might not be the best solution. My testing shows that with this the connection is being gracefully shutdown.

In file platform.c, function socket_close(sock_p s), line 2056 I added the following:

```
struct linger sl;
sl.l_onoff = 1;
sl.l_linger = 1;
[...]

Issue #35, opened 2016-10-26
Multi-threaded Capabilities 


kieranHay:
Thanks in advance for taking the time to read this.

I have multiple AB PLCs that I create, connect, read & write on different threads so that they run in parallel. However when a PLC is offline I try to reconnect/create the tags again until it succeeds. My question is, should this affect the other PLCs that are connected and are reading/writing?

At the moment I am having issues that it seems to be blocking the other PLC threads.

Thanks again.

kyle-github:
Hi,

Sorry for the delay.  I was trying to get version 1.5 out.

Are you using hostnames or IP addresses?  Hostnames require DNS look ups and those were fairly buggy in verison 1.0.  They are better in 1.5.  

The connect call is generally blocking, so if a PLC goes offline and you try to connect to it, a number of things could happen.  If the PLC just went offline, then it is possible that the network in between the application PC and the PLC has not realized this yet and you may hit timeouts.  If it has been a while (OS and setting dependent I believe), then it should fail quickly.  But, this should only effect one thread at a time.

However, if you are using different threads for every tag (I am not sure if that is what you were saying), then you should not have an overall hang.  I wonder if there is a mutex contention problem internally in the library?    

Can you provide more information about how your application uses the library so that I can try to duplicate this?

kyle-github:
I have been looking at this.  I think that you've stumbled on a mutex contention problem.

Here are the events that look like they will block in this case:

* create tags 1 and 2.
* do operations on both tags for a while.
* tag 2's PLC goes offline and tag 2 is closed.
* the application tries to reopen tag 2.
* the tag creation process continues fine until it gets to the point of opening the socket to PLC 2.  But PLC2 is not there.  In some circumstances, the network may not fail immediately to connect.  But, during this time, the tag creation has a central mutex locked.
* at the time this is happening, the other tag tries to read.  Part of that process involves queuing a request packet against the session object that is being created.  That involves locking the same mutex as in the step above.

This seems like a bug that needs to be fixed.  Unfortunately, it is not a simple one as the logic flow in question will need to be broken up into multiple parts.  Thanks for bringing this to my attention!

kieranHay:
Sorry for not getting back to you. Was going to create a simple test project showing the issue (and make sure I didn't mess up) but never got round to it. However you have found the issue. 
Thanks for investigating, I have gotten around this for now by using multiple programs for each plc. 

If you would like me test anything or help in anyway, I'm always willing to. Thanks again!

[...]

PR #634, opened 2026-06-13
Remove SO_LINGER option

vl-80:
The `SO_LINGER` option, set with zero timeout, causes abrupt termination of TCP connections. See issue https://github.com/libplctag/libplctag/issues/631.

The `SO_LINGER` option was added in an attempt to solve a problem with mutex contention due to blocking behavior of `connect()` (see https://github.com/libplctag/libplctag/issues/35).

The enabling of this option was added in https://github.com/libplctag/libplctag/commit/b6eb51cd9b4a788a9fdbb67276c5e7998afa4873. Later commit https://github.com/libplctag/libplctag/commit/06e1ac340c8b6f5f262c1f3e5c2e20b55d8bcfe9 changed the linger timeout to zero, introducing the issue with abrupt TCP connections termination.

Then, commit https://github.com/libplctag/libplctag/commit/60d8926121112d7156bf8f0dfab70f635c14ad6a switched to the non-blocking socket behavior. It alone mitigated the original problem described in the issue https://github.com/libplctag/libplctag/issues/35. The `SO_LINGER` option, however, remained in the code. Currently, not only does it fail to serve any purpose, but it also causes abrupt termination of all TCP connections.


kyle-github:
This looks correct to me.  I did some research and setting non-blocking on the sockets should let the app close the socket on the PC end without blocking even if there is outstanding data to write or read.  Thanks!

kyle-github:
All tests pass.  Thank you and @mikecflood for finding this and providing a fix!

kyle-github:
This is now queued up for the next release.

Issue #638, opened 2026-06-16
Data race on the refcount cleanup-thread run flag


aleks-f:
The deferred-cleanup worker thread is gated by `static volatile int cleanup_thread_running`. It is read and written from different
threads with no synchronization:

- **read** by the worker thread as its loop condition in `refcount_cleanup_thread_func`   (`while(cleanup_thread_running)`), and by `rc_dec_impl` as the gate before queueing   deferred cleanup (runs on arbitrary caller threads);
- **written** by the main thread in `refcount_startup` (set to 1 before `thread_create`,   cleared to 0 on the create-error path) and in `refcount_teardown` (cleared to 0 at   shutdown, reached via `plc_tag_shutdown` -> `destroy_modules` -> `refcount_teardown`).

PR #640, opened 2026-06-16
638 thread cleanup race

aleks-f:
#638


kyle-github:
I am going to close this because most of it is against the wrong ab_server and as noted elsewhere, this is not how the plan is for server tags.

Issue #637, opened 2026-06-16
Build robustness: stack-direction check + PIC/TLS for shared-library use


aleks-f:
This probably belongs in the libyafl repo. We have our own threaded server that runs well, so it's not urgent/critical, but just to give a feedback/proposal.

Three issues building/embedding libyafl:

1. **Stack-direction check hard-fails when the probe can't run.** The CMake `try_run(check_stack_direction.c)` `FATAL_ERROR`s with "stack grows upward"  whenever it can't execute the probe (sandboxed configure, cross build without an emulator), even on downward-stack arches.

2. **The probe is wrong under optimization.** `check()` gets inlined into `main()` at `-O1`+, collapsing the two frames, so it returns "upward"
   (clang `-O1/-O2/-O3` -> exit 1).

3. **Can't link into a shared library.** `yafl.c`'s `_Thread_local  tls_current_fiber` uses the local-exec TLS model; the resulting
   `R_AARCH64_TLSLE` (etc.) relocations are rejected with `-shared`. (asm has no TLS, so no asm change is needed.)

**Fixes:**
- Probe: mark `check()` `noinline` and take the addresses through `volatile` (correct at all `-O`).
- Check: warn + assume downward instead of `FATAL_ERROR`.
- Build PIC when used in a shared lib.

Optional: gate the `test_yafl_*` targets behind `BUILD_TESTING` so vendoring projects don't inherit them (the by-design-segfault `test_yafl_guard` especially).

PR #641, opened 2026-06-16
fix(libyafl): make the stack-direction check robust #637

aleks-f:
Mark the probe noinline (and take frame addresses through volatile) so it is correct at -O1 and above, where the compiler otherwise inlines it into main and returns a bogus upward result. Make the CMake check advisory (warn and assume downward) instead of fatal when the probe cannot be executed (sandboxed configure, or a cross build without an emulator): every supported architecture grows the stack downward and asm availability is the real gate.

#637


kyle-github:
Is this written by AI?  The amount of text and language in the text seems similar to something AI might produce.

aleks-f:
> Is this written by AI? The amount of text and language in the text seems similar to something AI might produce.

AI was used to some degree in producing all of the pull requests sent recently, but I would not call it "written by AI"; the process is not just prompting and blindly applying the responses. All responses are human-reviewed and modified/tuned if needed, then built and ran on our main implementation platforms (linux and windows, x64 and arm64) as part of our codebase (which exposed most of the problems in the first place, especially the sanitized test runs). 

Currently, all changes pass our full asan/ubsan/tsan test run. But even sanitizers, especially tsan, may miss problems; so, for anything problematic not discovered during the testing process, I'm counting on libplctag CI to uncover it. 

In regards this particular PR, fiber server is nice and interesting, but not a must for us at this time - thread server works fine and more important is the capability to have clients and servers in the same process (which we still had to tweak to make 2.7 work). FWIW, our tests of the integrated fiber server on both linux and windows currently pass (after the last PR fixing the windows libplctag hang).

aleks-f:
CI fail was not from my changes - there was a pre-existing bug here:

```
WRITE of size 4 thread T2
    #0 compat_atomic_inc_int32                 tests/utils/compat_utils.c:137
    #1 tag_callback                            tests/test_fairness/test_fairness.c:96
    #2 plc_tag_generic_handle_event_callbacks  lib/lib.c
    #3 tag_tickler_func                        lib/lib.c
  freed by main                                tests/test_fairness/test_fairness.c:300
  ```
  
Instead of defaulting to asan only for debug builds, I recommend sanitized (asan, ubsan, tsan) CI jobs. Even release build sanitized runs are better than nothing, often even absolutely better because of faster execution, which increases the chance of hitting the problem. This bug was discovered after 22 repeated sanitized test runs. After the fix, 40 test runs passed. TSan would have likely flag it right away.

kyle-github:
It does look like it is a intermittent timing issue.  I am going to add CI runs with sanitizers, but I want to keep the production builds clean.  The sanitizers slow down certain parts and allocate extra memory.

aleks-f:
> It does look like it is a intermittent timing issue. I am going to add CI runs with sanitizers, but I want to keep the production builds clean. The sanitizers slow down certain parts and allocate extra memory.

Just to clarify: 
I was not proposing to ship sanitized binaries. The suggestion was to have automated CI sanitized builds/testruns of release-optimized code. Executing sanitized release binary has its benefits because you're examining code layout (and execution speed) eventually delivered to users ; in many scenarios, this leads to faster discovery of intermittently manifesting problems (especially threading-related ones). 

Once you have a problem manifest in CI release build, if the culprit is not immediately obvious from the sanitizer trace (adding debug info to release build helps, but optimized release code layout is never the same as in debug build), you can reproduce and pinpoint it locally with full debug build.

Issue #639, opened 2026-06-16
Data race: `plc_tag_create` calls vtable `status`/`abort` without `api_mutex`


aleks-f:
`plc_tag_create` calls `tag->vtable->status(tag)` - the post-create poll and again inside the create wait loop - and `tag->vtable->abort(tag)` on each create-error path, all without holding `tag->api_mutex`. Every other status caller takes the lock first; the unlocked create-path access is a data race.

For the AB protocol the status function `ab_tag_status` reads `read_in_progress`, `write_in_progress`, and `status`, while the tickler and the abort helper `ab_tag_abort_request_only` write them under `api_mutex`. TSan flags both the `read_in_progress` read (size 4) and the `status` read (size 1) against those writes. The same pattern applies to any protocol whose `status`/`abort` vtable entries touch tag fields the tickler updates.

**Fix:** wrap the create-path `vtable->status` and `vtable->abort` calls in `critical_block(tag->api_mutex)`, mirroring `plc_tag_status_impl`. `api_mutex` is recursive, so no self-deadlock, and only the bare call goes inside the block so no early return escapes the lock. This is a generic-layer fix (no protocol or struct/ABI change) and closes the race for all protocols at once.

PR #642, opened 2026-06-16
fix(lib): hold api_mutex for create-path tag-state access #639

aleks-f:
plc_tag_create touched tag state shared with the background tickler thread without holding tag->api_mutex: the two vtable->status calls, the three create-error vtable->abort calls (which reach ab_tag_status / the abort helper reading and writing read_in_progress, write_in_progress and status), and the post-timeout read_in_flight/write_in_flight clears plus the create event. The tickler mutates all of these under api_mutex, so each was a data race. Every other status caller already locks via plc_tag_status_impl. Wrap these create-path accesses in critical_block(tag->api_mutex); the mutex is recursive, so it is deadlock-safe.

#639


kyle-github:
Nice catch!

kyle-github:
This code shows that there needs to be a better way to ensure shared data is correctly updated.  While this works, it highlights that we should probably be doing more with atomics to avoid overuse of mutexes.

Issue #635, opened 2026-06-16
Data race on `terminating` / `abort_request` flags


aleks-f:
`ab_session.terminating` and `ab_request.abort_request` (and the Omron  equivalents) are shared between the API/tag threads (set in `ab_tag_abort*` / `omron_tag_abort*`) and the connection-handler thread (`destroy_session`, the handler loop, send/recv paths, `purge_aborted_requests_unsafe`, `request_destroy`), but are plain/`volatile int` accessed without a lock - a data race (TSan flags it).

Related: v2.7.0 added `session_mutex` around the `max_payload` readers (`session_get_max_payload`,  session_get_available_cip_payload_space`) but the writer assignments (`max_payload_guess`/`max_payload_size` in `try_forward_open` and the forward-open response handlers) are still unguarded.

spiel2001:
### Independent confirmation on v2.5.3 and v2.7.0, with a public reproducer and the specific unguarded accesses

We hit this in a larger application and then built a small standalone reproducer to
isolate it (see attached ZIP). It uses only the public API, needs nothing but the bundled `ab_server`,
and reports the race in about 20 seconds. It is CC0 — please fold it into the test
suite if it is useful.

#### How to reproduce

Build libplctag with ThreadSanitizer, then build the attached `tsan_repro.cpp`
against it. Then:

```bash
# terminal 1
ab_server --plc=ControlLogix --path=1,0 --tag=TestDint:DINT[16000]

# terminal 2
TSAN_OPTIONS="halt_on_error=0 history_size=7" \
  ./tsan_repro --tags=8 --seconds=20 --cycle-us=100
```

The process exits 66 (TSAN's default `exitcode`). Measured on a 4-core machine:

| Version | TSAN warnings | reports in `purge_aborted_requests_unsafe` | reports in `ab_tag_abort*` |
|---|---|---|---|
| v2.5.3 | 15 | 2 | 3 |
| v2.7.0 | 21 | 5 | 3 |

Both endpoints were the same version in each case — the client library and the
`ab_server` it talks to — so cross-version protocol mismatch is not a factor.

The `--tags` and `--cycle-us` values matter. At the reproducer's defaults
(`--tags=4`, `--cycle-us=2000`) v2.5.3 still reports the race but v2.7.0 usually
does not; eight tags with a 100 microsecond read-to-abort delay reports it on both.

#### What the reproducer does

One thread per tag, no tag shared between threads. Per cycle it issues an
[...]

PR #643, opened 2026-06-16
635 terminating abort request race

aleks-f:
#635


kyle-github:
We should think about removing the spin lock.  If everything is atomic, maybe we do not need it.

Issue #636, opened 2026-06-16
Make `ab_server_fiber` embeddable as a shared library (in-process server, #590)


aleks-f:
`src/poc/ab_server_fiber` is executable-only, so a client and the simulator
can't run in one process (#590). Proposal: add an opt-in shared-library build
plus two small protocol/usability gaps.

**Embeddable lib** (CMake `AB_SERVER_FIBER_LIB` option):
- Build a shared lib (distinct target, `OUTPUT_NAME ab_server`) exposing
  `ab_server_main(argc,argv)` / `ab_server_stop()` (already guarded by
  `AB_SERVER_LIB`), so a host runs it on its own thread.
- Export the entry points with default visibility (the lib builds fine with
  `-fvisibility=hidden`); gate the SIGINT handler + the stats dump out of the
  lib build so the host owns signals/stdout.

**ListIdentity (0x63):** add an EtherNet/IP ListIdentity responder returning a
CIP Identity object - neither `tools/ab_server` nor the fiber server answers
0x63 today, so embedders can't report controller identity.

**CLI tolerance:** skip-and-warn on unknown flags and accept a bare `--debug`,
so a host can pass through extra args without aborting startup.

Depends on the libyafl build fixes (separate issue) for the shared-lib build.

PR #644, opened 2026-06-16
636 embeddable ab server fiber

aleks-f:
#636

### DIFF
--- a/src/poc/ab_server_fiber/CMakeLists.txt
+++ b/src/poc/ab_server_fiber/CMakeLists.txt
@@ -58,3 +58,39 @@ elseif(UNIX)
     target_link_libraries(ab_server_fiber m)
     target_compile_options(ab_server_fiber PRIVATE -Wall -Wextra)
 endif()
+
+# ---- embeddable shared-library variant ----
+# Opt-in: build the simulator as a shared library exposing
+# ab_server_main()/ab_server_stop() (guarded by AB_SERVER_LIB in
+# ab_server_fiber.c) so a host process can run it in-process on its own thread.
+# OFF by default, so the standalone executable build is unaffected.
+option(AB_SERVER_FIBER_LIB "Also build the fiber ab_server as an embeddable shared library" OFF)
+if(AB_SERVER_FIBER_LIB)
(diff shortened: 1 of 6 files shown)


kyle-github:
Please break this up so that I can more clearly see what is going on. For instance, support for ListIdentity is a different set of changes from making the server part of the library.  Other things:

- ab_server_fiber is a proof-of-concept (poc), not a fully supported tool.  The tool ab_server is the fully supported one.

- please file all PRs against the `prerelease` branch.  I do not merge anything into the `release` branch that has not gone through testing in the `prerelease` branch.

- I am not sure that this is the right way to do this.  I would prefer that we had special tag types that became served types rather than have the server itself be a part of the library.  This will take more work, but will fit into the existing model better.

The ab server is going to get a lot of changes in the near future because I cannot use it to test ListIdentity, GetIdentity, tag and UDT listing, and full OMRON support.  

I am going to close this for now.  I appreciate the effort very much!  However, this is too much change in one PR and it is not in the direction I want to take for supporting server-like behavior in the library.  Please respond if you think this is a mistake and tell me your reasons.

kyle-github:
This is the sketch of how I am thinking about server support in the library:

```c
int32_t server_tag = plc_tag_create_ex("protocol=ab-eip-server&listen=127.0.0.1:44818&path=1,4&plc=ControlLogix&name=myArrayTag:DINT[2,3,4]&debug=4", server_callback, server_data, 1000);
```

This is at the brainstorming stage.  

- the protocol indicates that this is a server tag.
- the plc type shows the type of PLC, but needs some way to show the actual model.  For now it could be hardcoded to just L80 or something.
- the tag name provides the tag and shape/type.  This borrows the syntax from the ab_server program.
- the callback is called when data is written or read from a client.
- there needs to be a way to call a callback periodically so that app code can simulate changes in a tag value.
- there needs to be some way to figure out how to set up an end point for device scanning and more details for identity.

If you have more ideas, please open an issue so we can discuss them!

kyle-github:
Oh, and there needs to be a way, or a decision, about how to handle multiple connections to the server tags.

Issue #509, opened 2024-12-31
Make available data type for ControlLogix tags


Hector-ER:
Hello,
   I was studing your code, because I want to detect the data type of the AB tag. I saw that when the tag is initialized you perform a Read, and then, in the answer you have some bytes that inform the data type and size.  You use this info to know the size of the buffer and also I can get the size, but I think that it's not available for me (user of the library) any information about data type. Am I right?
Is it possible to make it available?

Best regards and Happy New Year.

Héctor.

timyhac:
libplctag v2.6.0 exposed this with [`plc_tag_get_byte_array_attribute` and the `raw_tag_type_bytes` array attribute](https://github.com/libplctag/libplctag/wiki/API#tag-internal-attribute-functions).

I note that you're using the libplctag.NET wrapper - there is an example of using the above [here](https://github.com/libplctag/libplctag.NET/blob/main/examples/CSharp%20DotNetCore/ExampleTagDynamic.cs#L20) but a warning that I have not tested this.

Hector-ER:
Thanks, this is exactly whay I was looking for.

There is a very small bug (in libplctag.NET).  In method GetByteArrayAttribute native plc_tag_get_byte_array_attribute function is called, and if it's ok it returns with the length of the buffer.  In the following line ThrowIfStatusNotOk(result) is called which will throw an exception with the result.  "libplctag.LibPlcTagException: '2'"
Except for this (in my very quick test) it's working fine.

timyhac:
I didn't realize that but I see from [here](https://github.com/libplctag/libplctag/blob/c55bc5876d938dda1c609750cde5ae4812d7b8a8/src/protocols/ab/ab_common.c#L936) that you are correct.

I will copy the [logic from tag_rw2](https://github.com/libplctag/libplctag/blob/c55bc5876d938dda1c609750cde5ae4812d7b8a8/src/examples/tag_rw2.c#L914) and check the return value is <0 before assuming it is an error.

Update: libplctag.NET v1.6.0-alpha.1 has been released with this change.

PR #645, opened 2026-06-17
feat(ab_server): embeddable in-process server for same-process client…

aleks-f:
…+server #509

- add ab_server_main()/ab_server_stop() (under AB_SERVER_LIB) so a host can run the simulator on a thread and stop it; export only these two entry points (hidden visibility) so ab_server and libplctag do not interpose each other's identically-named utility symbols when loaded in one process
- add EtherNet/IP ListIdentity (0x63) responder with a neutral, unspecified-vendor identity
- fix multi-instance shutdown: the shutdown flag was a function-local static, so concurrent in-process servers shared one object (data race + cross-instance shutdown); move it into struct tcp_server as a per-instance atomic_bool
- make the terminate flag atomic_int (set cross-thread by ab_server_stop, not just a signal) and drain in-flight connection handlers before tcp_server_start returns
- free the tag list on exit; handle SIGTERM/SIGQUIT

### DIFF
--- a/src/tools/ab_server/eip.c
+++ b/src/tools/ab_server/eip.c
@@ -45,6 +45,7 @@
 #define EIP_UNREGISTER_SESSION ((uint16_t)0x0066)
 #define EIP_UNCONNECTED_SEND ((uint16_t)0x006F)
 #define EIP_CONNECTED_SEND ((uint16_t)0x0070)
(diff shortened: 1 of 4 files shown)


kyle-github:
See the comments about the previous PR to do embedding of the server.  Please open an issue with ideas for discussion.  What are your use cases?

Issue #653, opened 2026-06-19
Add EtherNet/IP ListIdentity (0x63)


aleks-f:
Returns a (configurable at build time) generic controller identity (vendor / device type / product code / revision / serial / product name).

kyle-github:
I have some major extensions planned for ab_server:

- UDP support for ListIdentity
- Identity Object
- Rockwell tag and UDT listing
- OMRON tag and UDT listing

I am currently working on a generic EtherNet/IP client implementation to clean up the duplicate code in the AB and OMRON protocol implementations.  That will use the Identity object of each PLC to get the actual model and version.

PR #654, opened 2026-06-19
feat(ab_server): add EtherNet/IP ListIdentity (0x63) #653

aleks-f:


kyle-github:
I am going to hang on to this one for a bit.  I am creating a new version of the simulator that merges the best (in my opinion) ideas from ab_server and ab_server_fiber.  It is thread based but uses the arena and byte functions from the fiber server.  These do a lot to make stack use much smaller and to make packet construction cleaner, though slower.

PR #657, opened 2026-07-15
fix(ab/pccc): SLC I/O file-type codes (0x8c/0x8b -> 0x83/0x82)

dhruvkpatel:
The PCCC file-type byte sent for SLC input and output image files was
0x8c/0x8b.

I've tested use of this library to read from input image files on an
SLC 5/05 (1747-L551/C, EtherNet/IP), and in doing so, discovered that
reads from input image files always fail. Locally, by changing the I
file-type code to 0x83, reads succeed, and match the output of a
reference implementation, pycomm3.

See here: https://github.com/ottowayi/pycomm3/blob/v1.2.16/pycomm3/cip/pccc.py#L146-L147

### Note on documentation
The documentation isn't exactly clear on this. The DF1 reference manual
(1770-6.5.16, p. 7-17) lists 0x8b/0x8c as "output/input logical by
slot" for this command, and 0x80-0x83 as "reserved" - but on real
hardware, the flat word-offset reads this library sends only work with
0x83/0x82. This also resolves the existing FIXME on the output case
("Check if 0x82 is correct instead of 0x8b"), which this change
removes.
https://literature.rockwellautomation.com/idc/groups/literature/documents/rm/1770-rm516_-en-p.pdf

### Scope
Only slc_encode_address() puts file_type on the wire;
plc5_encode_address() encodes logical-binary levels with no type byte,
so PLC-5 requests are unaffected.

### Validation
Compared reads of all 93 slot-addressed inputs of an SLC 5/05 through
this library and through pycomm3, before and after the change. Before,
every input read was rejected, and after all reads succeeded and
matched the pycomm3 values. I've also tested on a MicroLogix 1766
(1766-L32AWA, embedded I/O): all 28 slot-addressed inputs read and
match pycomm3 with the new value.

### Note on Output byte
I have not validated the output byte (0x82) - I'm not yet able to
safely test that on real hardware. I'm proposing to change both because
it's very likely they're paired together and pycomm3 uses this value as
well. I can leave it out of this change if we'd rather wait for actual
hardware validation on that.


kyle-github:
This looks good.  I have a Micrologix.  I will try to read I and O.

kyle-github:
Hmm, looks like there is a regression.  I don't think it is in your code.  I'll dig into this.

dhruvkpatel:
thanks!

kyle-github:
I have applied the patch manually.  Should be out in a day or two. Thank you!!!

